PR #5, opened 2019-09-06
Use HTTPS Protocol

emiluzelac:
Just a minor HTTPS tweak.

### DIFF
--- a/header.php
+++ b/header.php
@@ -7,7 +7,7 @@
 		<meta http-equiv="content-type" content="<?php bloginfo( 'html_type' ); ?>" charset="<?php bloginfo( 'charset' ); ?>" />
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" >
 
-		<link rel="profile" href="http://gmpg.org/xfn/11">
+		<link rel="profile" href="https://gmpg.org/xfn/11">
 
 		<?php wp_head(); ?>
 


pattonwebz:
LGTM :+1:

joyously:
Why not remove the entire line? It's not actually needed these days.

header.php
```diff
@@ -132,4 +132,4 @@
 		<?php 
 		// Output the menu modal
 		get_template_part( 'parts/modal-menu' ); 
-		?>
\ No newline at end of file
+		?>
```
ZebulanStanphill: There's no reason to have this closing tag, and looking at Twenty Nineteen, it looks like the convention is to omit it at the end of a file.

emiluzelac:
@carolinan Sorry, I forgot to add GitHub/WordPress username for the [CONTRIBUTORS.md](https://github.com/WordPress/twentytwenty/blob/master/CONTRIBUTORS.md)

Both are emiluzelac

Thanks!

header.php
emiluzelac: @ZebulanStanphill I didn't add that, just changed the protocol :)

ZebulanStanphill: Oh, I see. The closing tag was already there, but your editor added a newline character.

emiluzelac: It's directly via GH, no idea :)

ZebulanStanphill: This repo has a `.editorconfig` that specifies having a line break at the end of every file, and apparently GitHub has native support for EditorConfig:

https://github.com/editorconfig/editorconfig.github.com/pull/48

So that's probably where it came from.

PR #11, opened 2019-09-06
Add coding standards

dingo-d:
This PR will add the initial coding standards. The ruleset is subject to change.

The default standards set are the theme review one (`WPThemeReview`) plus the `WordPress-Extra` and `WordPress-Docs` (Extra contains Core standards so it doesn't have to be included).

For ease of use, composer scripts are added. To check the theme one needs to type in the terminal:

```bash
composer lint .
```

To see the summary:

```bash
composer lint-summary .
```

If and when the `package.json` is added, `node_modules` should be added to the `.gitignore` list as well.

EDIT:

Added travis so people can see what is failing and why 🙂 

We can try to fix this in this PR, but I'd rather fix this in a separate one and leave this one to possible improvements.

Would love if someone from the @WordPress/wpcs team could look at the PR 🙂


phpcs.xml.dist
```diff
@@ -0,0 +1,32 @@
...
+  <!-- Include the WordPress ruleset, with space for exclusions if necessary. -->
+  <rule ref="WordPress-Extra">
+    <exclude name="Generic.WhiteSpace.ScopeIndent.Incorrect" />
+    <exclude name="Generic.WhiteSpace.ScopeIndent.IncorrectExact" />
+
+    <exclude name="PEAR.Functions.FunctionCallSignature.Indent" />
+
+    <exclude name="Squiz.Commenting.FileComment.SpacingAfterComment" />
+    <exclude name="Squiz.Commenting.FunctionComment.MissingParamTag" />
+    <exclude name="Squiz.Commenting.InlineComment.InvalidEndChar" />
```
pattonwebz: Are we excluding these 2 inline comment sniffs intentionally? I would prefer to enforce these as this is how other core inline comments are done: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#5-inline-comments

.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+  directories:
+    # Cache directory for older Composer versions.
+    - $HOME/.composer/cache/files
+    # Cache directory for more recent Composer versions.
+    - $HOME/.cache/composer/files
+
+language:
+  - php
+
+php:
```
pento: Don't need this section, as the PHP versions are in the `matrix`, instead. It's just causing each PHP version to be run twice on Travis.

.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+  - 5.6
+  - 7.0
+  - 7.1
+  - 7.2
+  - 7.3
+  - "7.4snapshot"
+
+matrix:
+  fast_finish: true
+  include:
```
pento: We don't need to run linting on every version of PHP, just run it on PHP 7.3, and add the `script` to the `matrix`.

Also, if this PR is just focussed on adding linting and Travis (but not fixing the linting problems), it should mark the linting job in `allowed_failures`.

.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+  allow_failures:
+    # Allow failures for unstable builds.
+    - php: "7.4snapshot"
+
+before_install:
+    - composer install --dev
+
+script:
+  # Lint the PHP files against parse errors.
+  - if find . -path ./vendor -prune -o -path ./bin -prune -o -name "*.php" -exec php -l {} \; | grep "^[Parse error|Fatal error]"; then exit 1; fi
```
pento: Is this really necessary? Won't PHPCS catch parse failures?

phpcs.xml.dist
```diff
@@ -0,0 +1,32 @@
...
+
+  <!-- Set a description for this ruleset. -->
+  <description>Coding standards used for checking the Twenty Twenty theme.</description>
+
+  <!-- Exclude patterns -->
+  <exclude-pattern>/vendor/*</exclude-pattern>
+  <exclude-pattern>/node_modules/*</exclude-pattern>
+
+  <!-- Include the WordPress ruleset, with space for exclusions if necessary. -->
+  <rule ref="WordPress-Extra">
```
pento: Rather than adding an ad-hoc set of exclusions, these should be added in followup PRs, with appropriate rationale.

phpcs.xml.dist
dingo-d: I think that `Squiz.Commenting.InlineComment.InvalidEndChar` sniff will be invoked if anything except `.`, `!` or `?` is used. I looked at what `twentynineteen` used and copied that over. Not 100% why this was added in the first place 🤔

.travis.yml
dingo-d: So basically, just keep the 7.3 version. Not sure I understand the 'add `script` to the `matrix`' part.

> Also, if this PR is just focussed on adding linting and Travis (but not fixing the linting problems), it should mark the linting job in allowed_failures.

Well, this is the tricky bit. I'd want to enforce the code to pass the PHPCS checks, at the same time, the current PR will clearly fail because I don't want to make changes and add the configuration in the same PR (I'd have tons of changed files that should be checked). If I set it as `allowed_failures` the PR could be merged, but it should be reverted on the other PR that will then fix the issues (so that code checks are included).

I thought that once the configuration is ok, it can be merged (despite it failing on Travis) and then a new PR will be made with just changes from the PHPCS fixes.

.travis.yml
dingo-d: To my knowledge, it doesn't. PHPCS just tokenizes the code line by line and then checks the sniffs against it. This is included in the WPCS travis probably to ensure that the PHP files don't contain any errors which would then affect the tokenization.

phpcs.xml.dist
dingo-d: Ok, will add the `Extra` and then excluded when needed and the situation demands it. I used the `twentynineteen` as a 'template' for what to use, and this was added there, so I thought it couldn't hurt 🙂

composer.json
```diff
@@ -0,0 +1,30 @@
...
+      "homepage": "https://github.com/WordPress/twentytwenty"
+    }
+  ],
+  "require-dev": {
+      "wptrt/wpthemereview": "^0.2.0",
+      "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0"
+  },
+  "scripts": {
+    "phpcs-i": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -i",
+    "lint": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -s",
```
jrfnl: ### Function

The script as-is won't work as the ruleset does not pass the `file`s, so anyone running this locally will get a "_No files to scan_" error.

Also: why pass `-s` to the script ? Either add it to the ruleset or leave it out completely.

### Naming of the scripts

`Linting` is something different from `Sniffing` the code. I'd recommend against calling the `phpcs` script `lint` to prevent confusion. Consider a name like `checkcs` instead.

For `lint-summary`, you could consider `checkcs-summary` (or not adding it at all, people can pass their own extra arguments to composer scripts anyway).

For `format`, an alternative name could be `fixcs`,

### Supported PHP version

Another thing to consider: the `WPThemeReview` ruleset includes `PHPCompatibility` with a `testVersion` of `5.2-`.
What is the minimum PHP version for `TwentyTwenty` ? If it is PHP 5.6, the `testVersion` will need to be overruled in the command-line command, so these composer scripts would be the place to do that.
_Note: The `testVersion` currently can not be overruled from a custom ruleset, this is a known issue in PHPCS._

It might also be helpful to make the minimum PHP version of the theme explicit by setting it in the `require` section of the `composer.json` file.

.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+
+  # Validate the composer.json file.
+  # @link https://getcomposer.org/doc/03-cli.md#validate
+  - composer validate --no-check-all --strict
+  - composer lint .
```
jrfnl: Make this script dependent on the `ENV` variable set in the matrix, like so:
`- if [[ "$PHPCS" == 1 ]]; then composer lint; fi`.

Move the `.` to the `phpcs.xml.dist` ruleset.

You don't need PHPCS to run against every single PHP version. All PHPCS sniffs are unit tested to give the same results across PHP versions, so you only need to run PHPCS on one PHP version - preferably a high one, such as PHP 7.3 as suggested above -.

No need to repeat the unit tests from PHPCS and the external standards themselves.

phpcs.xml.dist
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
+<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="WordPress Coding Standards" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/squizlabs/PHP_CodeSniffer/master/phpcs.xsd">
```
jrfnl:
```suggestion
<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="TwentyTwenty Coding Standards" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/squizlabs/PHP_CodeSniffer/master/phpcs.xsd">
```

phpcs.xml.dist
```diff
@@ -0,0 +1,32 @@
...
+    <exclude name="Squiz.Commenting.InlineComment.InvalidEndChar" />
+    <exclude name="Squiz.Commenting.InlineComment.NotCapital" />
+  </rule>
+  <rule ref="WordPress-Docs" />
+  <rule ref="WPThemeReview" />
+
+  <rule ref="Squiz.Commenting.FunctionComment.ScalarTypeHintMissing">
+    <severity>0</severity>
+  </rule>
+</ruleset>
```
jrfnl: You probably want to add the following:

```xml
<rule ref="WordPress.Files.FileName">
	<properties>
		<property name="is_theme" value="true"/>
	</properties>
</rule>
```
See: https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#themes-allow-filename-exceptions

phpcs.xml.dist
```diff
@@ -0,0 +1,32 @@
...
+
+    <exclude name="Squiz.Commenting.FileComment.SpacingAfterComment" />
+    <exclude name="Squiz.Commenting.FunctionComment.MissingParamTag" />
+    <exclude name="Squiz.Commenting.InlineComment.InvalidEndChar" />
+    <exclude name="Squiz.Commenting.InlineComment.NotCapital" />
+  </rule>
+  <rule ref="WordPress-Docs" />
+  <rule ref="WPThemeReview" />
+
+  <rule ref="Squiz.Commenting.FunctionComment.ScalarTypeHintMissing">
```
jrfnl: This block should be removed. WPCS already excludes this.

.gitignore
```diff
@@ -0,0 +1 @@
+/vendor
```
jrfnl: Please add the typical names for PHPCS ruleset overloads devs can use locally to `.gitignore`:
```
.phpcs.xml
phpcs.xml
```

You may also want to add a `.gitattributes` file to `export-ignore` certain files, such as the Travis script and the `phpcs.xml.dist` file from being included in the packaged release zip/download from Packagist.

.travis.yml
jrfnl: As for the workflow: I'd suggest opening the PR(s) fixing CS failures concurrently. Once those have been merged, this PR can be rebased and the build should then pass.

Pro-tip: Small PRs targetting specific CS failures are easier to review than having one large PR fixing everything in one go.
This also allows for discussions on which rules should or should not be excluded to be had separately instead of all in one go.

.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+    - php: 7.1
+    - php: 7.0
+    - php: 5.6
+
+  allow_failures:
+    # Allow failures for unstable builds.
+    - php: "7.4snapshot"
+
+before_install:
+    - composer install --dev
```
jrfnl:
```suggestion
    - composer install
```
The Composer default is to install `--dev` nowadays.

composer.json
```diff
@@ -0,0 +1,30 @@
...
+  "type": "package",
+  "license": "GPL-2.0-or-later",
+  "keywords": [
+    "WordPress",
+    "Themes"
+  ],
+  "authors": [
+    {
+      "name": "the WordPress team",
+      "homepage": "https://github.com/WordPress/twentytwenty"
```
jrfnl: Should this link to the `contributors` page on GH as it is supposed to be the _Homepage of the authors_ ?

You can add a top-level `homepage` item for the project homepage.

Refs:
* https://getcomposer.org/doc/04-schema.md#homepage
* https://getcomposer.org/doc/04-schema.md#authors

composer.json
```diff
@@ -0,0 +1,30 @@
...
+    }
+  ],
+  "require-dev": {
+      "wptrt/wpthemereview": "^0.2.0",
+      "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0"
+  },
+  "scripts": {
+    "phpcs-i": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -i",
```
jrfnl: This should not be needed as a script.

phpcs.xml.dist
```diff
@@ -0,0 +1,32 @@
...
+  <!-- See https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml -->
+  <!-- See https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/.phpcs.xml.dist -->
+
+  <!-- Set a description for this ruleset. -->
+  <description>Coding standards used for checking the Twenty Twenty theme.</description>
+
+  <!-- Exclude patterns -->
+  <exclude-pattern>/vendor/*</exclude-pattern>
+  <exclude-pattern>/node_modules/*</exclude-pattern>
+
```
jrfnl: * You're missing the `file` directive. The Composer script as-is in composer will error with a "No files to scan provided" error unless the `.` is explicitly passed (which Travis does, but shouldn't need to).
    ```xml
    <file>.</file>
    ``` 
* What about adding:
    - `basepath`
    - `parallel`
    - `cache`
    - and possibly `extensions`

phpcs.xml.dist
jrfnl: I concur with @pento. I'd start with excluding nothing. Twenty-Twenty should - IMO - encompass all best practices.

If, while fixing the issues, it turns out a particular rule is not realistic to include, it can be excluded later or selectively excluded, i.e. excluded just for a particular group of files, not for the whole theme.

It would also be good to include inline comments in the ruleset documenting why certain rules are excluded.

Also, for documentation purposes, it would be better to have exclude rules within the ruleset which added them, so if the last three (`Squiz.Commenting....`) would be kept, they should be moved to the `WordPress-Docs` ruleset.

phpcs.xml.dist
pattonwebz: hmmmm :thinking: yeah seems like it just slipped in unexpectedly there. Other comments look to have suggestions here anyways making my point moot. We can reevaluate later about it if needed :)

.travis.yml
dingo-d: Ok, so what I'll do is:
- Add PHP-Parallel-Lint to composer
- Add php versions from 5.6 to 7.3
- Lint all the versions but only check CS on PHP7.3 (when the `PHPCS=1`)
- Add `Generic.PHP.Syntax` exclusion to the `phpcs.xml.dist`

Thanks for the suggestions 🙂

composer.json
dingo-d: Thanks for the suggestions, I think I just copied over the command names from the [core](https://github.com/WordPress/wordpress-develop/blob/4bd76c15b5f8f311ba93e1bc653cad6f7c297881/composer.json#L18). The `checkcs` and `fixcs` makes more sense definitely 🙂 

I'll add the necessary arguments in the `phpcs.xml.dist` (except the `testVersion`, I'm aware of this issue).
I'll remove the summary, that can be added if needed, as you've mentioned. And I'll use the `lint` for actually linting the php files.

Thanks for the comment :+1:

phpcs.xml.dist
dingo-d: I like this idea. Since #35 aims to add linting and build process for JS and CSS, phpcs should just focus on PHP 🙂 :+1:

dingo-d:
I've made updates according to the suggestions, a big thanks to all for that 🙂 

The summary of the issues is

```bash
PHP CODE SNIFFER REPORT SUMMARY
----------------------------------------------------------------------
FILE                                                  ERRORS  WARNINGS
----------------------------------------------------------------------
404.php                                               2       1
comments.php                                          22      11
content.php                                           21      12
footer.php                                            23      18
functions.php                                         205     54
header.php                                            11      11
index.php                                             7       6
pagination.php                                        4       1
singular.php                                          3       0
template-cover.php                                    2       0
template-full-width.php                               2       0
parts/modal-menu.php                                  24      16
parts/modal-search.php                                2       0
parts/page-header.php                                 5       0
parts/classes/class-comment-walker.php                13      17
parts/classes/class-svg-icons.php                     25      19
parts/classes/class-theme-customizer.php              210     90
----------------------------------------------------------------------
A TOTAL OF 581 ERRORS AND 256 WARNINGS WERE FOUND IN 17 FILES
----------------------------------------------------------------------
PHPCBF CAN FIX 602 OF THESE SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

Time: 803ms; Memory: 16MB


PHP CODE SNIFFER VIOLATION SOURCE SUMMARY
---------------------------------------------------------------------------------------
    SOURCE                                                                        COUNT
---------------------------------------------------------------------------------------
[x] WordPress.WhiteSpace.DisallowInlineTabs.NonIndentTabsUsed                     206
[x] WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned             161
[ ] Squiz.Commenting.InlineComment.InvalidEndChar                                 102
[x] PEAR.Functions.FunctionCallSignature.CloseBracketLine                         50
[x] PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket                  50
[x] PEAR.Functions.FunctionCallSignature.MultipleArguments                        36
[x] Generic.Formatting.MultipleStatementAlignment.NotSameWarning                  28
[ ] Squiz.Commenting.FunctionComment.Missing                                      27
[ ] WordPress.Security.EscapeOutput.OutputNotEscaped                              20
[ ] WordPress.WhiteSpace.PrecisionAlignment.Found                                 19
[x] Generic.Files.EndFileNewline.NotFound                                         14
[x] Generic.WhiteSpace.DisallowSpaceIndent.SpacesUsed                             14
[x] Squiz.Commenting.BlockComment.NoNewLine                                       13
[x] Squiz.PHP.EmbeddedPhp.ContentBeforeEnd                                        13
[ ] WordPress.PHP.StrictInArray.MissingTrueStrict                                 12
[ ] Squiz.Commenting.FileComment.Missing                                          10
[ ] WordPress.PHP.StrictComparisons.LooseComparison                               10
[ ] WordPress.PHP.YodaConditions.NotYoda                                          8
[ ] Squiz.Commenting.FileComment.WrongStyle                                       6
[ ] Squiz.Commenting.ClassComment.Missing                                         4
[ ] Squiz.Commenting.FunctionComment.WrongStyle                                   4
[x] Generic.WhiteSpace.ScopeIndent.Incorrect                                      3
[ ] WordPress.Files.FileName.InvalidClassFileName                                 3
[x] PSR2.ControlStructures.ElseIfDeclaration.NotAllowed                           2
[x] Squiz.ControlStructures.ControlSignature.NewlineAfterOpenBrace                2
[x] WordPress.Arrays.CommaAfterArrayItem.NoComma                                  2
[ ] Generic.Files.OneObjectStructurePerFile.MultipleFound                         1
[x] Generic.Functions.FunctionCallArgumentSpacing.TooMuchSpaceAfterComma          1
[x] Generic.Functions.OpeningFunctionBraceKernighanRitchie.SpaceBeforeBrace       1
[x] Generic.WhiteSpace.ScopeIndent.IncorrectExact                                 1
[ ] PSR2.Classes.PropertyDeclaration.ScopeMissing                                 1
[ ] Squiz.Commenting.FunctionComment.SpacingAfter                                 1
[ ] Squiz.Commenting.VariableComment.WrongStyle                                   1
[x] Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword                    1
[x] Squiz.PHP.EmbeddedPhp.ContentAfterOpen                                        1
[ ] Squiz.PHP.NonExecutableCode.ReturnNotRequired                                 1
[ ] WPThemeReview.Templates.ReservedFileNamePrefix.ReservedTemplatePrefixFound    1
[ ] WordPress.Security.EscapeOutput.UnsafePrintingFunction                        1
[ ] WordPress.WP.EnqueuedResourceParameters.NotInFooter                           1
[ ] WordPress.WP.GlobalVariablesOverride.Prohibited                               1
[ ] WordPress.WP.I18n.NonSingularStringLiteralText                                1
[x] WordPress.WhiteSpace.ControlStructureSpacing.ExtraSpaceAfterCloseParenthesis  1
[x] WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterStructureOpen        1
[x] WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceBeforeOpenParenthesis     1
---------------------------------------------------------------------------------------
A TOTAL OF 837 SNIFF VIOLATIONS WERE FOUND IN 44 SOURCES
---------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 22 MARKED SOURCES AUTOMATICALLY (602 VIOLATIONS IN TOTAL)
---------------------------------------------------------------------------------------
```

I'll start working on fixes for these issues (I've already started to work on those). The plan is to fix all the big issues that can be fixed automatically (tabs, alignments, etc.), then move on others that need a manual inspection.

phpcs.xml.dist
jrfnl: If I remember correctly, there are actually some WPThemeReview sniffs which examine CSS and possibly JS as well, so be careful with that. All the same, making explicit that either only PHP files are scanned or that all three extensions are scanned, is not a bad thing in itself.

.travis.yml
```diff
@@ -0,0 +1,40 @@
...
+    - composer install
+
+script:
+  # Validate the composer.json file.
+  # @link https://getcomposer.org/doc/03-cli.md#validate
+  - composer validate --no-check-all --strict
+
+  # Lint the PHP files against parse errors.
+  - composer lint
+  - if [[ "$PHPCS" == 1 ]]; then composer checkcs; fi
```
jrfnl: This should probably have its own comment:
```yaml
# Check the code against the WPCS code standards and best practices, as well as for compliance with the Theme Review rules and PHP cross-version compatibility.
```
or a shorter alternative:
```yaml
# Check the code against the standards as documented in the phpcs.xml.dist file.
```

composer.json
```diff
@@ -0,0 +1,34 @@
...
+  "require": {
+    "php": ">=5.6"
+  },
+  "require-dev": {
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
+    "jakub-onderka/php-parallel-lint": "^1.0",
+    "wptrt/wpthemereview": "^0.2.0"
+  },
+  "scripts": {
+    "lint": "@php ./vendor/bin/parallel-lint --exclude app --exclude vendor .",
```
jrfnl: Pro-tip: add `--exclude .git` to this.

I've done some testing with this application and while it won't find PHP files in the `.git` directory, if not explicitly excluded, it will still scan the directory, making it the scan slower.

Also: what's the `--exclude app` for ?

composer.json
```diff
@@ -0,0 +1,34 @@
...
+    "checkcs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs --runtime-set testVersion 5.6-",
+    "fixcs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf --report=summary,source --runtime-set testVersion 5.6-"
```
jrfnl:
```suggestion
    "fixcs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf --runtime-set testVersion 5.6-"
```

Not sure if you added the `report` on purpose here, but `phpcbf` has its own report and won't display other reports when run.

composer.json
dingo-d: Oh must be a copy-paste error 😬 I'll fix this and add the `.git` folder :+1:

phpcs.xml.dist
ntwb: A follow up PR should make switch this to PHP files please @dingo-d

PR #17, opened 2019-09-06
Move wp_body_open right after <body> tag

### LINKED ISSUE #16: wp_body_open should be triggered right after <body> tag
`wp_body_open` is expected to be triggered right after `<body>` opening tag and before any HTML bit of content.

audrasjb:
`wp_body_open` is expected to be triggered right after `<body>` opening tag and before any HTML bit of content.

Fixes #16

### DIFF
--- a/header.php
+++ b/header.php
@@ -15,13 +15,13 @@
 
 	<body <?php body_class(); ?>>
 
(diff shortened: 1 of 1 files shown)


andersnoren:
I included this as part of #36, which also includes @pattonwebz suggestion in #32.

Issue #10, opened 2019-09-06
comment references Twenty Nineteen


joyously:
In functions.php line 92:

https://github.com/WordPress/twentytwenty/blob/81b787b65c3e1a19c7e230cf4b49729ebd970760/functions.php#L92-L93

carolinan:
Merged (props already added for audrasjb and needs to be added for joyously).

PR #18, opened 2019-09-06
Fix/remove twentynineteen reference

audrasjb:
Replace Twenty Nineteen reference with Twenty Twelve in `functions.php`.


audrasjb:
Fixes #10

senadir:
I can't see any changes to the PR

audrasjb:
Thanks @senadir that was a commit issue :)

Issue #9, opened 2019-09-06
Remove the byline/date on pages


kraftbj:
On a static page, I'd love to not have the author or the date listed. It doesn't really matter that my about page is 9 years old (screenshot below).

At least as an option if there's a desired decision to keep it.
![2019-09-06 at 16 08](https://user-images.githubusercontent.com/88897/64460691-add18b00-d0c0-11e9-9815-9cc84abb37e3.jpg)

joyously:
Yes, that information should not be shown on static Pages. It's part of the definition of the post type.

carolinan:
Fixed and merged with https://github.com/WordPress/twentytwenty/pull/20

PR #20, opened 2019-09-06
Fix post meta output on pages

andersnoren:
Updates the `twentytwenty_get_post_meta` function to check the current post type before outputting post meta. Only `post` outputs post meta by default, but the `twentytwenty_allowed_post_types_for_meta_output` filter can be used by child themes and plugins to set which post types should be able to output post meta.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -518,9 +518,9 @@ if ( ! function_exists( 'twentytwenty_get_post_meta' ) ) :
 
 		$page_template = get_page_template_slug( $post_id );
 
-		// Check that the post type should be able to output post meta
-		$allowed_post_types = apply_filters( 'twentytwenty_allowed_post_types_for_meta_output', array( 'post' ) );
-		if ( ! in_array( get_post_type( $post_id ), $allowed_post_types ) ) {
+		// Check whether the post type is allowed to output post meta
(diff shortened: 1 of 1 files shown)


andersnoren:
Fixes #9

joyously:
Can you invert that? So all custom post types don't have to be added? And attachments work also?
If 'page' is the default for types that don't want output, then it's easier to filter to add other types to limit output, instead of having to add all custom types to get the output.

andersnoren:
Good point. I've updated the PR.

functions.php
```diff
@@ -518,9 +518,9 @@ function twentytwenty_get_post_meta( $post_id = null, $location = 'single-top' )
 
 		$page_template = get_page_template_slug( $post_id );
 
-		// Check that the post type should be able to output post meta
-		$allowed_post_types = apply_filters( 'twentytwenty_allowed_post_types_for_meta_output', array( 'post' ) );
-		if ( ! in_array( get_post_type( $post_id ), $allowed_post_types ) ) {
+		// Check whether the post type is allowed to output post meta
```
carolinan: Inline comments must end in full-stops, exclamation marks, or question marks.
https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#5-inline-comments

PR #30, opened 2019-09-06
FIX #29: Add `async` and `defer`, async main script

### LINKED ISSUE #29: `async` JavaScript, add ability to `async` and `defer`
The enqueued JavaScript should be set to `async` to improve performance and prevent render blocking. Core ticket [12009](https://core.trac.wordpress.org/ticket/12009) extends `wp_scripts_add_data` to allow `async` and `defer` as options. WP Rig [has a solution](https://github.com/wprig/wprig/blob/master/inc/Base_Support/Component.php#L145-L173) which gracefully steps back if/when core adopts this modern loading practice.

mor10:
- Adds `async` and `defer` as options to control JS loading via `wp_script_add_data`. 
- Sets `/assets/js/construct.js` to `async` to improve performance and reduce render blocking.

### DIFF
--- a/parts/classes/class-script-loader.php
+++ b/parts/classes/class-script-loader.php
@@ -1,0 +1,39 @@
+<?php
+
+/* ---------------------------------------------------------------------------------------------
+   JAVASCRIPT LOADER CLASS
+   Allow `async` and `defer` while enqueueing JavaScript. Based on a solution in WP Rig. 
+   --------------------------------------------------------------------------------------------- */
+
+class TwentyTwenty_Script_Loader {
+
+	/**
+	 * Adds async/defer attributes to enqueued / registered scripts.
+	 *
+	 * If #12009 lands in WordPress, this function can no-op since it would be handled in core.
+	 *
+	 * @link https://core.trac.wordpress.org/ticket/12009
+	 *
+	 * @param string $tag    The script tag.
+	 * @param string $handle The script handle.
+	 * @return string Script HTML string.
+	 */
+	public function filter_script_loader_tag( string $tag, string $handle ) : string {
+		foreach ( [ 'async', 'defer' ] as $attr ) {
+			if ( ! wp_scripts()->get_data( $handle, $attr ) ) {
+				continue;
+			}
+			// Prevent adding attribute when already added in #12009.
+			if ( ! preg_match( ":\s$attr(=|>|\s):", $tag ) ) {
+				$tag = preg_replace( ':(?=></script>):', " $attr", $tag, 1 );
+			}
+			// Only allow async or defer, not both.
+			break;
+		}
+		return $tag;
+	}
+	
+}
+
+$loader = new TwentyTwenty_Script_Loader;
+add_filter( 'script_loader_tag', [ $loader, 'filter_script_loader_tag' ], 10, 2 );
--- a/functions.php
+++ b/functions.php
@@ -118,6 +118,9 @@ require get_template_directory() . '/parts/classes/class-theme-customizer.php';
 // Custom comment walker
 require get_template_directory() . '/parts/classes/class-comment-walker.php';
 
+// Handle JavaScript loading
+require get_template_directory() . '/parts/classes/class-script-loader.php';
+
 /**
  * Register and Enqueue Styles
  */
@@ -159,6 +162,7 @@ if ( ! function_exists( 'twentytwenty_register_scripts' ) ) :
 		$js_dependencies = array( 'jquery' );
 
 		wp_enqueue_script( 'twentytwenty-construct', get_template_directory_uri() . '/assets/js/construct.js', $js_dependencies, $theme_version );
+		wp_script_add_data( 'twentytwenty-construct', 'async', true );
 
 	}
 	add_action( 'wp_enqueue_scripts', 'twentytwenty_register_scripts' );


pattonwebz:
This is a neat trick :smile:

Think we instantiate the loader somewhere else though? Perhaps inside an `after_setup_theme` hook?

parts/classes/class-script-loader.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+/* ---------------------------------------------------------------------------------------------
```
dingo-d: Can we use [docblock](https://www.phpdoc.org/git) style comments instead these big huge blocks?

We should use standard PHP practices.

parts/classes/class-script-loader.php
```diff
@@ -0,0 +1,39 @@
...
+	 *
+	 * If #12009 lands in WordPress, this function can no-op since it would be handled in core.
+	 *
+	 * @link https://core.trac.wordpress.org/ticket/12009
+	 *
+	 * @param string $tag    The script tag.
+	 * @param string $handle The script handle.
+	 * @return string Script HTML string.
+	 */
+	public function filter_script_loader_tag( string $tag, string $handle ) : string {
```
dingo-d: Type hinting against `string` argument and return value will only work on PHP7.0, and the minimum requirement (at this point) of PHP in WordPress is still 5.6.

https://mlocati.github.io/articles/php-type-hinting.html

parts/classes/class-script-loader.php
```diff
@@ -0,0 +1,39 @@
...
+		foreach ( [ 'async', 'defer' ] as $attr ) {
+			if ( ! wp_scripts()->get_data( $handle, $attr ) ) {
+				continue;
+			}
+			// Prevent adding attribute when already added in #12009.
+			if ( ! preg_match( ":\s$attr(=|>|\s):", $tag ) ) {
```
dingo-d: Would `stripos` to match if an attribute exists in the tag work like regex match? Regex matching is usually a bit slower than matching against a string.

parts/classes/class-script-loader.php
```diff
@@ -0,0 +1,39 @@
...
+			}
+			// Only allow async or defer, not both.
+			break;
+		}
+		return $tag;
+	}
+	
+}
+
+$loader = new TwentyTwenty_Script_Loader;
```
dingo-d: Since this theme is not build using proper OOP principles (encapsulation, polymorphism etc.), I see no reason to put this code in a separate class. There is no dependency injection used anywhere so we can just write procedural code.

parts/classes/class-script-loader.php
mor10: Afaik this format is the expected format for WordPress coding standards. It's the same format used in previous default themes.

parts/classes/class-script-loader.php
mor10: Keeping it as a class makes it nicely isolated and easy to remove if/when core finally allows for `async` and `defer`. `functions.php` is already large in this theme. If functionality got split up more, I'd agree to move it, but as of right now it seems more rational to keep it as a stand-alone class.

parts/classes/class-script-loader.php
mor10: Cc @felixarntz and @westonruter

ntwb: I agree, this theme needs to support PHP 5.6

Type hinting will have to wait for the PHP 7.0 bump

parts/classes/class-script-loader.php
ntwb: See https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers

parts/classes/class-script-loader.php
```diff
@@ -0,0 +1,39 @@
...
+	 * @param string $tag    The script tag.
+	 * @param string $handle The script handle.
+	 * @return string Script HTML string.
+	 */
+	public function filter_script_loader_tag( string $tag, string $handle ) : string {
+		foreach ( [ 'async', 'defer' ] as $attr ) {
+			if ( ! wp_scripts()->get_data( $handle, $attr ) ) {
+				continue;
+			}
+			// Prevent adding attribute when already added in #12009.
```
ntwb: Please update the link on a line of its own to point to the actual Trac ticket

`@link https://core.trac.wordpress.org/ticket/12009`

parts/classes/class-script-loader.php
dingo-d: > Afaik this format is the expected format for WordPress coding standards. It's the same format used in previous default themes.

Yeah, previous themes had some issues that passed through because TRT wasn't involved 😄We're trying to polish this one out so that we can say: look at 2020 🙂

parts/classes/class-script-loader.php
dingo-d: I understand the rationale, but it just mixes class-based with procedural code, which is a bit inconsistent. I'm all for keeping certain functionalities in separate files and including them in the `functions.php` for easier management, but I see no benefit in having a class with just one public method in it.

pattonwebz: I am actually of the oppinion that encapsulation, even just for the sake of it, is fine. I do not like including a class file that has side effects though.

felixarntz: I agree with @dingo-d here that, given how the rest the theme is coded, we shouldn't break that paradigm by introducing a class for functionality like this, as that can make the code harded to understand than everything following one approach.

I suggest introducing a single function attached to the filter.

pattonwebz:
We talked about this and we want to roll this in. I hope it encourages others to follow this newer best practice :)

I still vote we keep it as a class if only for the sake of ease of reuse/removed when this lands in core.

pattonwebz:
Hey @mor10 I grabbed this branch and made a couple adjustments to it. Moved/renamed the class to be beside others, moved the instantiation an filter to be alongside other theme support type items. It would be great if I could get a little feedback from you on it. 

The branch I've got that in is here: https://github.com/pattonwebz/twentytwenty/tree/issue/29-2/async-defer-javascript - it's the 5 most recent commits in this branch that are relevant

ianbelanger79:
Closing because this was merged in #339

PR #34, opened 2019-09-07
FIX #19: Add ESLint and stylelint with WP coding standards configured.

### LINKED ISSUE #19: Consider adding config for linting tools to ensure consistency
Add config for ESLint, stylelint, PHPCS aligned to WordPress Coding Standards to ensure code consistency during development.

mor10:
Adds basic framework for linting of JavaScript and CSS:  

- [ESLint](https://www.npmjs.com/package/eslint) with [`@wordpress/eslint-plugin`](https://www.npmjs.com/package/@wordpress/eslint-plugin)
- [stylelint](https://www.npmjs.com/package/stylelint) with [`stylelint-config-wordpress`](https://www.npmjs.com/package/stylelint-config-wordpress)

To use, either employ CLI commands or add ESLint and stylelint extensions to the code editor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "twentytwenty",
+  "version": "1.0.0",
+  "description": "Default WP Theme",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wordpress/twentytwenty.git"
+  },
+  "license": "GPL-2.0+",
+  "bugs": {
+    "url": "https://github.com/wordpress/twentytwenty/issues"
+  },
+  "homepage": "https://github.com/wordpress/twentytwenty#readme",
+  "devDependencies": {
+    "@wordpress/eslint-plugin": "^3.0.0",
+    "eslint": "^6.3.0",
+    "stylelint": "^10.1.0",
+    "stylelint-config-wordpress": "^14.0.0"
+  },
+  "eslintConfig": {
+    "extends": "plugin:@wordpress/eslint-plugin/recommended",
+    "root": true,
+    "env": {
+      "browser": true,
+      "node": true,
+      "es6": true
+    },
+    "globals": {
+      "jQuery": "readonly",
+      "wp": "readonly"
+    }
+  },
+  "stylelint": {
+    "extends": "stylelint-config-wordpress"
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "homepage": "https://github.com/wordpress/twentytwenty#readme",
+  "devDependencies": {
+    "@wordpress/eslint-plugin": "^3.0.0",
+    "eslint": "^6.3.0",
+    "stylelint": "^10.1.0",
+    "stylelint-config-wordpress": "^14.0.0"
+  },
+  "eslintConfig": {
+    "extends": "plugin:@wordpress/eslint-plugin/recommended",
```
senadir: This will add rulesets for React & jsx, something we don't use, we might change it to es5.
```suggestion
    "extends": "plugin:@wordpress/eslint-plugin/es5",
```

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "root": true,
+    "env": {
+      "browser": true,
+      "node": true,
```
senadir: as far as I'm aware all the js code will run on the browser, so we don't need to include node environment since it let some errors go by, also if we're supporting es6 then we should include it in the eslint extend
```suggestion
      "browser": true,
      "node": true,
      "es6": true
```

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@wordpress/eslint-plugin": "^3.0.0",
+    "eslint": "^6.3.0",
+    "stylelint": "^10.1.0",
+    "stylelint-config-wordpress": "^14.0.0"
+  },
+  "eslintConfig": {
+    "extends": "plugin:@wordpress/eslint-plugin/recommended",
+    "root": true,
+    "env": {
+      "browser": true,
```
senadir: as far as I'm aware all the js code will run on the browser, so we don't need to include node environment since it let some errors go by, also if we're supporting es6 then we should include it in the eslint extend
```suggestion
      "browser": true,
      "node": true,
      "es6": true
```

package.json
mor10: We should use ES6 as that is the current forward-thinking JS model. ES5 is an old and dying standard.

ntwb:
I'd love to see the `@wordpress/scripts` package used here instead 🙇

jeffpaul:
@mor10 thoughts on @senadir's comments and @ntwb's recommendation on alternatively using the scripts package?

ianbelanger79:
Closing this because more recent PR's take care of the linting

Issue #32, opened 2019-09-07
Use `wp_body_open` for output of bypass block


pattonwebz:
We should move the skip link to be output via the `wp_body_open` filter instead of coding it directly to the header.

This is the method that I would like to go with for all default themes. We have a ticket ready for commit on twentyeleven for reference: https://core.trac.wordpress.org/ticket/47891

This method of adding skip links avoid the issues which are becoming common in themes that #16 issue notes and #17 resolves.

joyously:
If you do it that way, it should have a higher priority than the default of 10. I mean bigger number so that it is after whatever is added to the hook with the default.

joyously:
I personally would rather see it in the template than in a filter. It makes it easier for users to grok.
My theme has an option to not show the header, so it has to conditionally output the skip-link. I think that's easier to see with it straight in the template file.

joyously:
I think the theme outputs all the HTML, so the new hook is for plugins to get in front with whatever (probably for AMP). It doesn't make sense to me that the theme needs to put anything on that hook.

ernilambar:
Agree with @joyously here. This hook is not supposed to be used to inject HTML like we are doing now. 
 
Ref - https://make.wordpress.org/core/2019/04/24/miscellaneous-developer-updates-in-5-2/

**Usage of this hook should be reserved for output of unseen elements like <script> tags or additional metadata. It should not be used to add arbitrary HTML content to a page that could break layouts or lead to unexpected situations.**

pattonwebz:
I remember writing that guideline... Didn't think it would come back to bite me :p

This is an initally unseen element, added to a location it is convenient via a hook. I think this is fine.

PR #36, opened 2019-09-07
Use `wp_body_open` for output of skip link

andersnoren:
Moves the output of the skip link from `header.php` to the `wp_body_open` action, matching the approach that will be used [in Twenty Eleven](https://core.trac.wordpress.org/ticket/47891). This also makes sure that `wp_body_open()` is triggered right after the `<body>`.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -328,6 +328,14 @@ if ( ! function_exists( 'twentytwenty_get_custom_logo' ) ) :
 	}
 endif;
 
+/**
+ * Shim for wp_body_open, ensuring backwards compatibility with versions of WordPress older than 5.2.
+ */
+if ( ! function_exists( 'wp_body_open' ) ) {
+	function wp_body_open() {
+		do_action( 'wp_body_open' );
+	}
+}
 
 /**
  * Include a skip to content link at the top of the page so that users can bypass the menu.
--- a/functions.php
+++ b/functions.php
@@ -328,6 +328,18 @@ if ( ! function_exists( 'twentytwenty_get_custom_logo' ) ) :
 	}
 endif;
 
+
+/**
+ * Include a skip to content link at the top of the page so that users can bypass the menu.
+ */
+if ( ! function_exists( 'twentytwenty_skip_link' ) ) :
+	function twentytwenty_skip_link() {
+		echo '<a class="skip-link faux-button" href="#site-content">' . esc_html__( 'Skip to the content', 'twentytwenty' ) . '</a>';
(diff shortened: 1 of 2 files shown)


andersnoren:
Fixed #32, #16

functions.php
```diff
@@ -328,6 +328,18 @@ function twentytwenty_get_custom_logo( $logo_theme_mod = 'custom_logo' ) {
...
 
+
+/**
+ * Include a skip to content link at the top of the page so that users can bypass the menu.
+ */
+if ( ! function_exists( 'twentytwenty_skip_link' ) ) :
+	function twentytwenty_skip_link() {
+		echo '<a class="skip-link faux-button" href="#site-content">' . esc_html__( 'Skip to the content', 'twentytwenty' ) . '</a>';
+	}
+	add_action( 'wp_body_open', 'twentytwenty_skip_link', 5 );
```
joyously: Shouldn't the priority be something larger than the default of 10?
The whole point of the new hook was for plugins that had to be first.

senadir: do you mean that plugins need to render before this link or after it?

joyously: I think before it.

aeonwp: I thought the idea was to keep the skip link before whatever plugins put there.

carolinan: I thought the idea was that the skip link needs to be before whatever plugins output there.

joyously: Did you read the ticket for that new hook? The plugins needed it so they could be first.

joyously: This confusion sounds like a good reason to leave the link in the template file and not use the hook at all.

pattonwebz:
Skip link should be before any other rendered content (so priority lower than default to prevent hook ordering issues). Content that does not render but wants to be first (aka Google tag manager) should use an even lower priority.

pattonwebz:
The PR looks correct to me in terms of how I imagine the prioritisation to work.

carolinan:
The link just needs to include the screen-reader-text class,then it can be merged.

carolinan:
Actually I believe we are also missing the fallback for wp_body_open?

andersnoren:
Oh right. I've updated the PR by removing the `function_exists()` conditional in header.php, and added the `wp_body_open()` shim suggested in https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/.

pattonwebz:
This looks good to me, fallback working fine on pre 5.2 :)

Issue #41, opened 2019-09-07
Include a license file


carolinan:
GPL asks that a copy of the license is included with the product.

While [it is suggested that a COPYING file](https://www.gnu.org/licenses/gpl-howto.html) should be used, WordPress itself includes a license.txt file and this is also the most common way for plugins and themes to include the full license.

Therefor a license.txt file with the GNU General Public License version 2.0 needs to be added.

PR #42, opened 2019-09-07
Create license.txt

carolinan:
Add a license.txt file with the GNU General Public License version 2.0.
Fixes https://github.com/WordPress/twentytwenty/issues/41


ntwb:
You should probably use the same license file as used by Gutenberg, this is the GPL 2 license and WordPress specifics in the single license file

PR #52, opened 2019-09-07
Editorconfig

Soean:
Adds an editorconfig file.


.editorconfig
```diff
@@ -0,0 +1,24 @@
...
+
+[{package.json,*.yml}]
+indent_style = space
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.txt]
+end_of_line = crlf
```
ntwb:
```suggestion
end_of_line = crlf

```

Adding an end-of-file new line, as all files require this per WP CS

Soean: Thanks for your review, I added the empty line.

PR #71, opened 2019-09-07
New header structure and style

### LINKED ISSUE #66: Replace all the href=# links with buttons
All the `href="#"` links and potentially other links with invalid URL or invalid page fragment identifiers need to be replaced with buttons:

WordPress Accessibility Coding Standards
Semantics for Controls
https://make.wordpress.org/core/handbook/best-practices/coding-standards/accessibility-coding-standards/#semantics-for-controls

Most notably:
- the link that opens the search
- the link that closes the search
- the link that opens the menu

Aside: once they're changed to `<button type="button" ...` there's no need to prevent the default action with `preventDefault()` or `return false`.

### LINKED ISSUE #68: Controls that disclose additional content need an `aria-expanded` attribute: remove `aria-pressed`
See https://github.com/WordPress/twentytwenty/issues/66

All controls that reveal additional content need to communicate users their purpose and state.

The proper attribute to use for this purpose is [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded). 

[aria-pressed](https://www.w3.org/TR/wai-aria-1.1/#aria-pressed) has a different purpose and needs to be removed in favor of `aria-expanded`.

andersnoren:
Updates the header structure and style to match the Twenty Twenty design. This commit also includes the new ”Shortcuts Menu”, which is displayed next to the header toggles on 1000px and up. We should have a discussion on what the most appropriate name for it is. I'll create an issue.

The commit also changes all toggles to be button elements instead of links (fixes #66), and changes the aria-pressed attribute to aria-expanded (fixes #68).

The commit does _not_ include accessible drop down menus for the shortcuts menu. I’m guessing we’ll use a JS based solution for that, in which case it might be better to wait until the jQuery → vanilla JS update is done.


functions.php
```diff
@@ -172,9 +172,10 @@ function twentytwenty_menus() {
 
         // Register menus
         $locations = array(
-            'main-menu' 	=> __( 'Main Menu', 'twentytwenty' ),
-            'footer-menu' 	=> __( 'Footer Menu', 'twentytwenty' ),
-            'social-menu' 	=> __( 'Social Menu', 'twentytwenty' ),
+            'footer-menu' 		=> __( 'Footer Menu', 'twentytwenty' ),
```
pattonwebz: One too many tabs?

andersnoren: That there was! Fixed now, thanks :)

andersnoren:
Fixed the merge conflict. The admin bar fix is working on my local.

carolinan:
When the screen size is reduced and the search is moved to the left of the site title, the visual order does not correspond with the semantic order.

The focus is now title, menu, and then all the way back to the left side for the search.
The expected result is the same as the visual order: search, title, menu.
https://make.wordpress.org/themes/handbook/review/accessibility/required/#keyboard-navigation

@afercia would be able to explain this better than me.

carolinan:
I believe that the two svg icons should not be focusable, because it means two tabs for each item, first the button and then the icon, and they both open the same modals.

pattonwebz:
I think adding `aria-hidden="true"`, `focusable="false"` would prevent the svgs being double tab to pass through. May also need `role="presentational"` just for more context to screen readers.

andersnoren:
Could be the `tabIndex` not having a capitalized `i`. Taking a look.

I'll have to think a bit on how to update the structure to match the focus order. We might need a different set of toggles for mobile/portrait tablet (when they are to left and right of the site title).

andersnoren:
@carolinan I've added the focus fix and updated the header structure to include mobile and tablet specific toggles, which are hidden → 1000px when the regular toggles appear.

afercia:
> When the screen size is reduced and the search is moved to the left of the site title, the visual order does not correspond with the semantic order.

Thanks @carolinan I just noticed this issue looking at the design mockups on [the Make post](https://make.wordpress.org/core/2019/09/06/introducing-twenty-twenty/), which show a different design from what's in the theme at the moment. I was unaware the header design was going to change.

header order

Can't test this PR at the moment but looking at the screenshot above, the order changes from large to small screens. 

If the order is changed via CSS, that's an accessibility problem. Reading and focus order must always match visual order.

Relevant WCAG requirements:
1.3.2 Meaningful Sequence (Level A) https://www.w3.org/TR/WCAG20/#content-structure-separation-sequence
2.4.3 Focus Order (Level A) https://www.w3.org/TR/WCAG20/#navigation-mechanisms-focus-order

Icon-only controls are also a big problem. Will create a separate issue later.

PR #97, opened 2019-09-07
ADD #91: Add print.css

### LINKED ISSUE #91: Add basic print styles
When I use Print Preview, it shows down to the page title on a blank page, then the content, navigation links, comments, comment form, and widget areas on subsequent pages. This all takes many pages.

There should be some print styles to reduce the font size, hide the interactive elements such as menu, nav links, comment form, widget areas, and get the content on the same print page as the title. 

print-page-1

print-page-2

nielslange:
Fixes #91


print.css
```diff
@@ -0,0 +1,118 @@
...
+    page-break-before: avoid;
+  }
+  /* Links */
+  a:link,
+  a:visited,
+  a {
+    background: transparent;
+    font-weight: bold;
+    text-decoration: underline;
+    text-align: left;
```
joyously: This might not be a good idea for RTL.
and links are bold _and_ underlined?

nielslange: @joyously I've copied and adjusted the print.css from the Twenty Nineteen theme and you definitely have a good point regarding `text-align: left;` for RTL. 

Regarding the link, don't you think it's better to have them formatted bold and underlined to highlight that those were links?

joyously: Whatever.... It seemed excessive to me since there is also the `href`.

print.css
```diff
@@ -0,0 +1,118 @@
...
+  .comment-form,
+  .comments-wrapper,
+  .comment .comment-metadata {
+    display: none;
+  }
+  .entry-content .wp-block-button .wp-block-button__link,
+  .entry-content .wp-block-button .wp-block-file__button,
+  .entry-content .button {
+    color: #000;
+    background: none;
```
joyously: I think there's a case of the title and post meta being white, for the cover image. That would be invisible for print.

print.css
```diff
@@ -0,0 +1,118 @@
...
+    content: "";
+  }
+  article a[href^="#"]:after {
+    content: "";
+  }
+  a:not(:local-link):after {
+    content: " < " attr(href) "> ";
+  }
+  /* Visibility */
+  .header-navigation-wrapper,
```
joyously: Are these classes right for this theme, or still what was copied from Twenty Nineteen?

nielslange: @joyously The class `.header-navigation-wrapper` is used in the `header.php` file.

joyously: Fine, I was actually asking about the whole list that started with that line.

carolinan:
Hi, 
we would like this to be moved to the style.css file (added using print media query).

Please hide:
the _responsive_ search and menu icons,
edit link
any post & archive pagination/navigation
comment form

nielslange:
@carolinan Doesn’t merging the print.css into the style.css cause a potential performance issue as the style.css gets bigger? I’m happy to merge them though, but I thought the print.css file would only get loaded when a user initiates printing the screen.

joyously:
>  I thought the print.css file would only get loaded when a user initiates printing the screen.

I thought so also. Can you use Developer Tools Network tab, to see if the browser loads the file?

joyously:
@carolinan The RTL label isn't needed here. The one CSS rule that would affect it should not even be there.

print.css
joyously: Definitely remove `text-align: left`

carolinan:
> @carolinan The RTL label isn't needed here. The one CSS rule that would affect it should not even be there.

The problem has not been solved as far as I can tell.

carolinan:
> @carolinan Doesn’t merging the print.css into the style.css cause a potential performance issue as the style.css gets bigger? I’m happy to merge them though, but I thought the print.css file would only get loaded when a user initiates printing the screen.

I dont enough about how big the performance hit would be.
(It would also be one more file or one less file to load.)  Only passing on information from discussions that has taken place. 

My personal comment would be that all scripts and styles must be enqued. There are more than one ways to do that with the print style.

print.css
nielslange: I removed `text-align: left` with https://github.com/WordPress/twentytwenty/pull/97/commits/1ef9a89fa44506f0a62a021efa5bb43fcc7331a0

carolinan:
Is there a reason to why it is not inside twentytwenty_register_styles?

nielslange:
@carolinan Due to performance reasons, I decided to load the `print.css` in the footer. Therefore, I cannot call the function within `wp_enqueue_scripts` but need to call it within `wp_footer`.

joyously:
Does calling wp_enqueue_style in wp_footer even work?
I don't think this is child theme friendly. 
And it definitely should have the main stylesheet as a dependency.

nielslange:
> Does calling wp_enqueue_style in wp_footer even work?
Yes, works without any problems. 

> I don't think this is child theme friendly.
What makes you think that, @joyously?

> And it definitely should have the main stylesheet as a dependency.
Why should the `print.css` have `style.css` as a dependency? I thought the whole purpose of having separate files is to style them differently.

joyously:
I don't think it's child theme friendly and it needs the dependency for the same reason: the order of the stylesheets, for the cascade. The print styles should come after the main stylesheet, and you could argue that by using wp_footer, they will, but then if the child theme wants to add some print styles, it would also have to use wp_footer. And any user styles added in Additional CSS would be in the head, so they couldn't override (they come out last in the head section).

I'm thinking that it would probably be better all around to put this at the bottom of style.css, or just enqueue it in the head with the main stylesheet as dependency.

> I thought the whole purpose of having separate files is to style them differently.

It is, but the print styles _do_ depend on the main styles. If you put the print styles first, they wouldn't work right.

nielslange:
> It is, but the print styles do depend on the main styles. If you put the print styles first, they wouldn't work right.

I wasn't aware of such a relationship yet and just run a quick test which confirmed your statement. Thanks for bringing this up and I've moved the function call for the `print.css` accordingly.

carolinan:
This is what I see in my print preview:

print

ianbelanger79:
We should probably wait until before RC to do a final check of the print styles. However, they could be iterated on as we go, to at least make them look better now.

joyously:
Yikes, that's worse than I've seen.

nielslange:
@carolinan Due to various changes some of the elements are out of synch. I'm currently working on an updated version, but as @ianbelanger79 mentioned, it probably makes more sense to finalise this issue shortly before the PR to ensure that the underlying structure does not change seriously.

PR #100, opened 2019-09-08
Add styles for post pagination

### LINKED ISSUE #99: Add styles to the newer / older post pagination
The newer / older post pagination is missing styles and does not yet match the initial screenshots.
Pagination.php should also be moved to the parts folder.

A decision needs to be made regarding https://github.com/WordPress/twentytwenty/issues/93 ,
_Icon-only controls_

In the mobile screenshot, only the arrow icons show and the text "Older /  Newer posts" is hidden.

If we choose to show both the icon and text, there is not enough room on small screens
to show the text, arrows and page numbers in one line.

Instead, the navigation could be split into 3 lines:

mobile-archive-pagination

And two lines if only one of the texts are displayed:

mobile-archive-pagination2

carolinan:
Add styles for the post pagination.
Moves pagination.php to the parts folder.
Code style changes.

Fixes #99


carolinan:
-This shows both icons and text on small screens.

andersnoren:
Another thought: if we use `flex-wrap: wrap`, we can remove the `display: block` from the `.prev` and `.next` so they only linebreak on mobile when they need to.

We could also add shorter versions of the text in the `.prev` and `.next` on mobile. So on page 1 on mobile, it would be:

image

carolinan:
So far:
The remaining top margin on the _next_ (older posts) doesn't look so good here. I'll keep testing a bit and then probably hand it over to someone else to continue :)
posts-nav

carolinan:
The translation strings for Newer and Older Posts now contains HTML, so that a shorter text can be shown on small screens. We should consider if there are any better solutions for those strings.

The position of the dots needs some work too.

style.css
```diff
@@ -2646,6 +2646,44 @@ div.comment:first-of-type {
 /*	13. Site Pagination
 /* --------------------------------------------------------------------------------------------- */
 
+.pagination .nav-links {
+	display: flex;
+	flex-wrap: wrap;
+	font-size: 1.8rem;
+	padding-top: 3rem;
+	font-weight: 600;
+}
```
andersnoren: Moving the top padding of `.nav-links` to the wrapping `.navigation.pagination` element, and adding `margin: -2.5rem 0 0 -2.5rem` and the equivalent positive values to the children of the `.nav-links` will allow us to have vertical margins between lines without any excess margins, and horizontal margins between items without an initial left margin on the first element of each wrapping line (as in the image below).

image

With this, you can use the follow targeting and CSS to get the margin between the child elements, reducing the overall amount of CSS a bit:

`.nav-links > * {
    margin: 2.5rem 0 0 2.5rem;
}`

This would also require the equivalent changes in the media query CSS.

Edit: Oh, and this will reduce the horizontal margin between the items a bit as well and remove the righthand margin from the last item in a row, slightly reducing the risk for line breaks altogether :)

style.css
```diff
@@ -2646,6 +2646,44 @@ div.comment:first-of-type {
 /*	13. Site Pagination
 /* --------------------------------------------------------------------------------------------- */
 
+.pagination .nav-links {
+	display: flex;
+	flex-wrap: wrap;
+	font-size: 1.8rem;
+	padding-top: 3rem;
+	font-weight: 600;
```
andersnoren: The `font-weight` value should be a `700` here.

style.css
```diff
@@ -2646,6 +2646,44 @@ div.comment:first-of-type {
...
+	text-decoration: underline;
+}
+
+.pagination .current {
+	margin: 0 2rem 2rem 2rem;
+}
+
+.pagination .dots {
+	display: inline-block;
+	vertical-align: middle;
```
andersnoren: Removing the `display: inline-block;`and `vertical-align: middle;`, and moving the `margin-top: -.3em` CSS that you have in the media query CSS to the base, seems to position the ellipsis correctly in all screen sizes for me.

style.css
```diff
@@ -3772,6 +3816,54 @@ ul.footer-social li {
 		margin: 1.2rem 0 0 2.4rem;
 	}
 
+	/* SITE PAGINATION */
```
andersnoren: There's a bigger comment heading called Site Pagination a couple of lines up in the CSS file which you can move this block into, for consistency.

style.css
```diff
@@ -2646,6 +2646,44 @@ div.comment:first-of-type {
 /*	13. Site Pagination
```
andersnoren: The pagination breaks a bit in IE 11, overflowing the right side of the screen. It could be related to `justify-self` and/or `display: initial`, which isn't supported in IE.

style.css
```diff
@@ -2646,6 +2621,38 @@ div.comment:first-of-type {
...
+.pagination a {
+	text-decoration: none;
+}
+
+.pagination a:hover {
+	text-decoration: underline;
+}
+
+.pagination .dots {
+	margin-top: 1.1em;
```
andersnoren: Changing this to `transform: translateY( -.3em );` should put it back in the center of the line. Other than that, the pagination looks good to me. :)

pattonwebz: Is this `em`s or should it be `rem`s?

Issue #12, opened 2019-09-06
Can we remove jQuery dependency?


joyously:
Would it be possible to rewrite the script in vanilla JS?

https://github.com/WordPress/twentytwenty/blob/81b787b65c3e1a19c7e230cf4b49729ebd970760/functions.php#L159

carolinan:
The plan is to remove jQuery yes.

LittleBigThing:
This seems to be a fair amount of work, but of course needed.

Is there a plan on how to do this? Maybe we can start summarising the JS driven stuff in this issue and split them out in separate tasks?

- [ ] Mobile menu toggle
- [ ] Search toggle
- [ ] Smooth scroll if needed
- [ ] ...

Also, we could start the modifications based on the _s theme's [vanilla JS files](https://github.com/Automattic/_s/tree/master/js).

LittleBigThing:
Some related issues:

- #27 
- #66 
- #68 
- #80 
- #81

PR #101, opened 2019-09-08
Move from jQuery to using Vanilla JS (ES6+)

fabiankaegy:
I tried to convert all the functionality currently stored in the 800+ line construct.js to a modern vanilla javascript syntax.

At the moment there are no frameworks or libraries used. 

In order to make the code more maintainable I've also split things up into their own components.

Everything is build using the [@wordpress/scripts](https://www.npmjs.com/package/@wordpress/scripts) package. 

This 100% needs some testing and somebody needs to decide wether its ok to have build tools for the javascript. And at the moment everything is just converted to a vanilla js approach. I didn't change anything about how for example the modals are handled. The logic and architecture stayed the same. Only the syntax and tooling is changed. 

#12


fabiankaegy:
#116 is also fixed in here

0aveRyan:
Think it's much better to put a dependency burden on a developer trying to contribute to the JS than for every site visitor with jQuery. These are also internal WordPress dependencies that can keep the focus on feature priorities.

karmatosed:
Thanks for exploring this. I would add a point that having older than WordPress 5.0 as a version shouldn't mean someone can't use the theme. If there can be some graceful fallback so menus and other elements still work, that would be the ideal outcome. It shouldn't prevent awesome refactoring like this, but let's consider and test that experience.

fabiankaegy:
@karmatosed yes ver much so. Thats why we decided in the core-theme chat on Monday to go down the route of #163 :)

carolinan:
Closing in favor of https://github.com/WordPress/twentytwenty/pull/163

PR #107, opened 2019-09-08
[WPCS] Fix inline tabs in spacing issue

dingo-d:
This PR is a part of the fix for the #11 issue. It addresses the `DisallowInlineTabsSniff` which checks if spaces are used for mid-line alignments ([handbook rule](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation)).

The fix is done using the automatic fixer (phpcbf).


functions.php
```diff
@@ -172,10 +172,10 @@ function twentytwenty_menus() {
 
         // Register menus
         $locations = array(
-            'footer-menu' 		=> __( 'Footer Menu', 'twentytwenty' ),
-            'main-menu' 		=> __( 'Main Menu', 'twentytwenty' ),
-			'shortcuts-menu' 	=> __( 'Shortcuts Menu', 'twentytwenty' ),
-            'social-menu' 		=> __( 'Social Menu', 'twentytwenty' ),
+            'footer-menu'       => __( 'Footer Menu', 'twentytwenty' ),
+            'main-menu'         => __( 'Main Menu', 'twentytwenty' ),
+			'shortcuts-menu'    => __( 'Shortcuts Menu', 'twentytwenty' ),
```
carolinan: The shortcuts-menu has too many tabs.

dingo-d: Good catch. Must be a mixed tabs/spaces issue. The fix is on the way 🙂

PR #112, opened 2019-09-08
[WPCS] Fix function signature issues

dingo-d:
This PR is a part of the fix for the #11 issue.
When splitting a function call over multiple lines, each parameter must be on a separate line ([handbook rule](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#multiline-function-calls)).

The fix is done using the automatic fixer (phpcbf).

Once other PRs are merged this will probably need to be revisited, from the quick glance it looks like some indentation is not ideal.


carolinan:
The tabs and spaces are incorrect again.

dingo-d:
> Once other PRs are merged this will probably need to be revisited, from the quick glance it looks like some indentation is not ideal.

Will need to merge master and make a new fix 🙂

footer.php
```diff
@@ -18,12 +18,15 @@
...
 							<?php
-							wp_nav_menu( array(
-								'container'      => '',
-								'depth'          => 1,
-								'items_wrap'     => '%3$s',
-								'theme_location' => 'footer-menu',
-							) );
+
+							wp_nav_menu(
+								array(
```
ntwb: You should be able to `wp_nav_menu( array(`

See https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation

I think the `array(` is part of PHPCBF automatically adding it to its own line which is not required, having `array(` on its own line ends up a little too much whitespace.

Denis can you update this instance, and if it passes the PHPCS check update all other instances of this too please

dingo-d: Putting the array after the function call triggers the `PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket` sniff, so this and `PEAR.Functions.FunctionCallSignature.CloseBracketLine` are good candidates for exclusions in the `phpcs.xml.dist`

pattonwebz: Agreed that this could result in a lot of whitespace that does not lend itself any better to readability than having the array on the function line.

dingo-d: We could just allow this: https://github.com/WordPress/WordPress-Coding-Standards/blob/ecb633d8cddb5fa38dc2de5d72b220810dde549c/WordPress-Core/ruleset.xml#L246

dingo-d:
I've submitted #175 that will address these issues. After this is done I need to look through the codebase and see if this PR is ok or not.

dingo-d:
Closing this one because we've ignored the fixes proposed by this PR

PR #115, opened 2019-09-08
Added missing classes output

andersnoren:
Fixes missing output of $featured_image_classes.


content.php
```diff
@@ -71,7 +71,7 @@
 
 			<figure class="featured-media">
 
-				<div class="featured-media-inner section-inner<?php esc_attr( $featured_media_inner_classes ); ?>">
+				<div class="featured-media-inner section-inner<?php echo esc_attr( $featured_media_inner_classes ); ?>">
```
jrfnl: Shouldn't this be `esc_attr_e()` instead ?

Issue #127, opened 2019-09-08
use proper hook for body_class


joyously:
`body_class` is a filter, not an action.
https://github.com/WordPress/twentytwenty/blob/37d7eddf5817f15b67e9707841c889a78d2b29de/functions.php#L255

carolinan:
Merged, thank you.

PR #138, opened 2019-09-09
Use proper hook for body_class

ernilambar:
`body_class` is a filter, not action.

---

WordPress.org ID: rabmalin


ernilambar:
Fixes https://github.com/WordPress/twentytwenty/issues/127

PR #143, opened 2019-09-09
Sanitize hex color used in theme mods

### LINKED ISSUE #130: Theme mod needs to be escaped
https://github.com/WordPress/twentytwenty/blob/e781eea060b3f02541ea6917e6df3ac531acd23f/functions.php#L911-L916

audrasjb:
In functions.php, use `sanitize_hex_color()` to add proper escaping to theme mods colors.
Fixes #130


acosmin:
`sanitize_hex_color()` is available for front-end only in WP 4.6 and higher... in case we need to take this into consideration

PR #146, opened 2019-09-09
Fixed: Use get_the_ID() insted of $post->ID

mukeshpanchal27:
Use get_the_ID() insted of $post->ID

<hr>

WordPress.org ID: mukesh27

### DIFF
--- a/parts/page-header.php
+++ b/parts/page-header.php
@@ -37,7 +37,7 @@
 		} else {
 
 			// Default to displaying the post meta
-			twentytwenty_the_post_meta( $post->ID, 'single-top' );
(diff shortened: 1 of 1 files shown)


carolinan:
Is this related to any current open issue? Please remember to include insights to why a change is needed.

mukeshpanchal27:
@carolinan We have to use **get_the_ID()** function within the WordPress loop to get current post id as per WP standard.

If we want to use **$post->ID** then we have defined **global $post** variable in the template.

carolinan:
Hi, this file has moved, do you have time to update the PR?

Issue #139, opened 2019-09-09
 Undefined variable notice for block and classic editors


acosmin:
https://github.com/WordPress/twentytwenty/blob/master/functions.php#L933

This will trigger a php Notice because it's defined only in that if statement and it's used down the line in two more if statements (where it's not defined).

You'll need to move it somewhere outside the if statements.

PR #153, opened 2019-09-09
Fixes undefined variable notice

mukeshpanchal27:
Fixes: #139 

<hr>

WordPress.org ID: mukesh27


carolinan:
I tried to move the variable up to be listed together with the others, but the functions.php file still needs to be updated.

PR #162, opened 2019-09-09
Replace endif's and : with braces

ianbelanger79:
Also added more space inside functions


ntwb:
> There are still 4 endif inside the twentytwenty_get_customizer_css.

This can be followed up in another PR, merging this whilst it can be withour further conflicts

PR #163, opened 2019-09-09
Move from jQuery to using Vanilla JS (ES5)

fabiankaegy:
After our discussion today in the core-themes channel in slack I took the work done in #101 and refactored it to only use es5 syntax so there is no need for build tooling. 

I still very much think that the other approach might be more maintainable but can see why build tooling might scare many away. So this should be all the work done over there without using any build tooling or modern js.


assets/js/index.js
```diff
@@ -0,0 +1,901 @@
...
+		falseVal = false;
+	}
+	if (element[attribute] !== trueVal) {
+		element.setAttribute(attribute, trueVal);
+	} else {
+		element.setAttribute(attribute, falseVal);
+	}
+}
+
+function getQueryStringValue() {
```
acosmin: Most of these functions should be prefixed (from bellow), they are in the the global scope

functions.php
```diff
@@ -154,9 +154,9 @@ function twentytwenty_register_scripts() {
 			wp_enqueue_script( 'comment-reply' );
 		}
 
-		$js_dependencies = array( 'jquery' );
+		$js_dependencies = array( 'wp-polyfill', 'wp-dom-ready' );
```
acosmin: I might be wrong but polyfill and dom-ready are available only in WP v5.*, everything below that will cause issues.

assets/js/index.js
```diff
@@ -0,0 +1,901 @@
...
+	init: function () {
+		this.makeFit();
+
+		window.addEventListener('fit-videos', function () {
+			this.makeFit();
+		});
+
+		window.addEventListener('resize', function () {
+			this.makeFit();
+		});
```
acosmin: There's a scope issue here, it should be something like:
```
window.addEventListener('resize', function () {
    this.makeFit();
}.bind(this));
```
or just use `this.makeFit` as the callback

andersnoren: @carolinan also suggested that `add_theme_support( 'responsive-embeds' );` would allow us to remove this block altogether.

acosmin: `responsive-embeds` is available only in WP 5.*. Shouldn't it be backwards compatible?

andersnoren: The theme review rules only allow us to be backwards compatible three major versions, which will be 5.0 when 5.3 is out.

acosmin: I guess so :) I thought this being a core theme should go even further...

acosmin: It's going to be set as a Featured theme and it will be in top 3 popular themes. Some of the user might find it that way and use it on older versions of WP

andersnoren: `responsive-embeds` was used in Twenty Nineteen as well, so there's some precedent for it. But I see your point.

assets/js/index.js
```diff
@@ -0,0 +1,901 @@
...
+			if (element.href === '#' || element.href === '#0' || element.classList.contains('.do-not-scroll') || element.classList.contains('skip-link')) {
+				return false;
+			}
+			return true;
+		}).forEach(function (element) {
+			element.addEventListener('click', function (event) {
+				// On-page links
+				if (window.location.hostname === event.target.hostname) {
+					// Figure out element to scroll to
+					debugger;
```
acosmin: This should be removed and to make the `event.target.hostname` work we should use some CSS on the specific buttons, that's if we don't want go more deep and add some parentNodes in the mix. The simple solution would be (eg: for back to top button):
```
a.to-the-top > * {
    pointer-events: none;
}
```

assets/js/index.js
```diff
@@ -0,0 +1,900 @@
...
+						event.preventDefault();
+
+						// Get options
+						var additionalOffset = event.target.dataset.additionalOffset;
+						var scrollSpeed = event.target.dataset.scrollSpeed ? event.target.dataset.scrollSpeed : 500;
+
+						// Determine offset
+						var originalOffset = target.getBoundingClientRect().top + window.pageYOffset;
+						var scrollOffset = additionalOffset ? originalOffset + additionalOffset : originalOffset;
+
```
acosmin: You could add this in case the adminbar is active:
```
var adminBar = document.querySelector( '#wpadminbar' );

if( adminBar && event.target.className === 'to-the-top' ) {
    scrollOffset = scrollOffset - adminBar.getBoundingClientRect().height;
}
```

pattonwebz:
@fabiankaegy scanning the code it looks good, I see no jQuery which was the main thing :)

I am not able to get the modals to trigger though for menu or the search. There were some markup adjustments there in recent days so maybe I need to change a target in the JS to get it working here again to test?

fabiankaegy:
@pattonwebz I just came across that aswell, when I merged all the changes. I will go over it again :)

assets/js/index.js
```diff
@@ -0,0 +1,906 @@
...
+--------------------------------------------------------------------------------------------------- */
+
+twentytwenty.intrinsicRatioVideos = {
+
+	init: function () {
+		this.makeFit();
+
+		window.addEventListener('fit-videos', function () {
+			this.makeFit();
+		}).bind(this);
```
acosmin: This should be `}.bind(this));`, same for the one below

fabiankaegy: just fixed that  👍 thanks :)

fabiankaegy:
@pattonwebz had a typo in one of the dependencies... the script did not load in the latest commit I had up here... Should be working again now :)

acosmin:
@fabiankaegy you should also remove the 'jquery' dependency 😄 that's the purpose of this PR, joking

fabiankaegy:
@acosmin Yes I had that removed. I thought that it should be in there for the purpose of "fixing" the admin bar until core fixes that problem. (There was some discussion about this last weeks core-theme chat) (But I agree that iim should be removed here and if needed queued elsewhere :))

acosmin:
The `twentytwenty.scrollLock` is broken, it keeps the `style="position: fixed; width: 100%; left: 0px; top: 0px;" when you untoggle the search modal making the page unscrollable... I think there's an issue with the `prevLockStyles` but I need to go out and don't have the time to investigate it right now 😄 

Also, I think this window attached proprieties should be prefixed.

fabiankaegy:
@acosmin hmmm that is wired, I'm not able to reproduce the locked styles not being removed :O 

Agreed on the prefixing :)

pattonwebz:
Can confirm that the modals now appear when clicked :)

I do also see the thing @acosmin mentions where the scroll lock remains in place when the search is closed.

fabiankaegy:
@pattonwebz how do you click out of the modal? just mouse click or using the X or using the escape key?

pattonwebz:
@fabiankaegy seems to happen for me all ways of existing. Esc key, `x` icon or clicking outside it. Looks like this CSS is being added after it closes onto the `html` element `position: fixed; width: 100%; left: 0px; top: 0px;`.

I am using Chrome on Ubuntu: 
Version 74.0.3729.157 (Official Build) (64-bit)

fabiankaegy:
@pattonwebz okay will investigate that. At the moment all are working for me on Safari 12.1.2 on macOS 10.14.6.

thanks for all the insight

fabiankaegy:
Okay, can conform that it is happening on Firefox and Chrome. Somehow is working in Safari. Will see on what's going on.

fabiankaegy:
@pattonwebz @acosmin the cross browser issue should be fixed in https://github.com/WordPress/twentytwenty/pull/163/commits/8d07123237dd2a9981ef08e1e875ca8af4c13c9e

It was something to do with the iteratebility of the `CSSStyleDeclaration`. So I now manually convert it into an object and re apply that object once the modal is removed. Tested on Firefox 68.0.1 (64-bit) Chrome 76.0.3809.132 and Safari 12.1.2 on macOS 10.14.6

fabiankaegy:
@pattonwebz @acosmin the cross browser issue should be fixed in https://github.com/WordPress/twentytwenty/pull/163/commits/8d07123237dd2a9981ef08e1e875ca8af4c13c9e

It was something to do with the iteratebility of the `CSSStyleDeclaration`. So I now manually convert it into an object and re apply that object once the modal is removed. Tested on Firefox Chrome 76.0.3809.132 and Safari 12.1.2 on macOS 10.14.6

pattonwebz:
Yup the scroll locking seems to be resolved here now on Chrome 74.

I am not certain what else would be needing before this could be mergable. If anyone with JavaScript experience can look at it that'd be good. It looks ok to me but surely someone that knows more JS than me can leave some comments.

assets/js/index.js
```diff
@@ -0,0 +1,914 @@
...
+		var anchorElementsList = Array.prototype.slice.call(anchorElements);
+		anchorElementsList.filter(function (element) {
+			if (element.href === '#' || element.href === '#0' || element.classList.contains('.do-not-scroll') || element.classList.contains('skip-link')) {
+				return false;
+			}
+			return true;
+		}).forEach(function (element) {
+			element.addEventListener('click', function (event) {
+				// On-page links
+				if (window.location.hostname === event.target.hostname) {
```
acosmin: The `event.target.hostname` here is `undefined` because the target is the `span.to-the-top-long` tag. The event should be on the `a` tag, it would work as intended if you add `pointer-events: none` on both span tags wrapped by this element.

assets/js/index.js
```diff
@@ -0,0 +1,914 @@
...
+					// Figure out element to scroll to
+					var target = window.location.hash !== '' && document.querySelector(window.location.hash);
+					target = target ? target : event.target.hash !== '' && document.querySelector('#' + event.target.hash);
```
acosmin: The `event.target.hash` here will always have a `#` so the `'#' +` part isn't needed.

assets/js/index.js
```diff
@@ -0,0 +1,914 @@
...
+
+		window.addEventListener('resize', function () {
+			this.setScreenHeight();
+		}.bind(this));
+	},
+
+	setScreenHeight: function () {
+		var screenHeight = document.querySelector('.screen-height');
+		if (screenHeight) {
+			screenHeight.style.minHeight = window.innerHeight;
```
acosmin: This doesn't work, it should be `window.innerHeight + 'px'`

Also, setTimeout has any purpose in `init()`?

assets/js/index.js
```diff
@@ -0,0 +1,914 @@
...
+		}
+	},
+
+}; // twentytwenty.smoothScroll
+
+/*	-----------------------------------------------------------------------------------------------
+	Interval Scroll
+--------------------------------------------------------------------------------------------------- */
+
+twentytwenty.intervalScroll = {
```
acosmin: I think we can get rid of this function, the event isn't used anywhere...

acosmin: same for `twentytwenty.resizeEnd`

fabiankaegy: agreed  👍 this far I only mirrored all the functionality from the original file. I did not jet take unused stuff out. Wasn't sure wether that should happen as part of the same PR.

acosmin:
@pattonwebz  I think it's ready for merge :) 

Maybe you can get someone else to take a look.

We could've probably made it compatible with WP versions under 5.0... the `dom-ready` dep can be written in 2-3 lines, as for `polyfill` :) I can't say how many of those are used...

@fabiankaegy great job on this!

fabiankaegy:
@acosmin cool thank you. I'm not opposed to replacing dom-ready with our own function. Just wasn't sure wether we really need write it again and again when it's already there. But yes it is only in 5.0 and up... 

as for the polyfill I am not certain myself maybe someone who is a bit more familiar with the package can say more about wether they think it really gives us a lot in regards to older browsers.

dingo-d:
One thing that I'm on the fence about is that all the JS code is in one file (it's 900+ loc). Could a component based approach be used here? So that all the individual parts (`coverModals`, `intrinsicRatioVideos`, `intervalScroll` etc.) are in their files as modules, and then the `init` is done in an `index.js` file using ES5 `require`?

This is what I'm doing (to an extent) in the theme sniffer. The [index.js](https://github.com/WPTRT/theme-sniffer/blob/development/assets/dev/scripts/index.js) file is where I define the parts which interact with the DOM, and the functionality is in the other class file. Granted I'm using ES6 because of the build tools, and I need to further extract certain functionality, but the principle is there.

fabiankaegy:
@dingo-d Hi there, yeah I first went with that approach, but then we decided in the [weekly chat last week](https://make.wordpress.org/core/2019/09/10/core-themes-chat-summary-for-september-4-2019/) that there should be no build process and therefore I removed all ES6. The first approach was #101

andersnoren:
@fabiankaegy First, huge thanks for taking the lead on this one – it's been a big relief for us that you've been able to put in so much work on it.

I thought I'd weigh in on what code I think can be removed altogether from the JavaScript file (since I'm responsible for the original jQuery code).

I believe these ones can be safely removed altogether:

```
twentytwenty.intervalScroll.init();			// Check for scroll on an interval
twentytwenty.dynamicScreenHeight.init();	// Dynamic Screen Height
twentytwenty.resizeEnd.init();				// Trigger event at end of resize
```

Since all of the modal toggles are in the top of the page, the scroll lock function can be rewritten to just set the `scrollTop` position to 0 and the body to `overflow: hidden`. This will also fix an issue we have with the admin bar position when the `scrollLock` (as it works currently) is active.

```
twentytwenty.scrollLock.init();				// Scroll Lock
```

I think we need some variation of the rest, but they can all be simplified significantly to be more specific to the elements in Twenty Twenty. 

There are also additions needed (handling focus for the horizontal menu, for example), but it might be better to make those changes once this PR is merged.

dingo-d:
> @dingo-d Hi there, yeah I first went with that approach, but then we decided in the [weekly chat last week](https://make.wordpress.org/core/2019/09/10/core-themes-chat-summary-for-september-4-2019/) that there should be no build process and therefore I removed all ES6. The first approach was #101

Oh so not even using basic require.js, that's sad :/

acosmin:
@andersnoren `Object.keys()` is supported in IE11, are you sure you didn't check it with IE10?

As for IE10, WP dropped support for it, so I wouldn't worry about it 😄

ianbelanger79:
@fabiankaegy Just tested the PR on Windows 10 in Firefox, Chrome, Edge and IE11.

Everything looks good, except the modals do not work in IE11. Other than that, I think this is in good shape.

functions.php
```diff
@@ -207,7 +207,8 @@ function twentytwenty_register_scripts() {
 
 		$js_dependencies = array( 'wp-dom-ready', 'wp-polyfill' );
 
-		wp_enqueue_script( 'twentytwenty-index', get_template_directory_uri() . '/assets/js/index.js', $js_dependencies, $theme_version, false );
+		wp_enqueue_script( 'twentytwenty-construct', get_template_directory_uri() . '/assets/js/construct.js', $js_dependencies, $theme_version, false );
```
fabiankaegy: This is now pointing to the wrong file. Should I rename the index.js file to construct.js again?

ianbelanger79: No I will fix it, I prefer index.js

ianbelanger79:
Unfortunately this still does not work in IE11

assets/js/index.js
```diff
@@ -0,0 +1,847 @@
...
+
+				if (targetString === 'next') {
+					target = toggle.nextSibling;
+				} else {
+					target = document.querySelector(targetString);
+				}
+
+				// Trigger events on the toggle targets before they are toggled
+				if (target.classList.contains('active')) {
+					target.dispatchEvent(new Event('toggle-target-before-active'));
```
acosmin: Seems like `Event()` isn't supported by IE

https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

😞

acosmin: I've tested this [polyfill ](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/Event/polyfill.js) and the menu and search modal work as intended...

fabiankaegy: I did not see the problem with the event coming. I will look into it and see wether there is a polyfill for it or wether we can rework the code to kit need the dispatching of custom events

fabiankaegy: I think I found a solution for it. I sadly do not have access to a machine running ie11 so my testing is limited. If you have the time I would love to get some feedback wether it is working now :)

fabiankaegy:
@ianbelanger79  I think I found a solution for the new Event problem. I sadly do not have access to a machine running ie11 so my testing is limited. If you have the time I would love to get some feedback wether it is working now :)

acosmin:
Eureka, it works 😸

acosmin:
One last issue we need to solve is `To the top ↑` doesn't work 😄 

Chrome latest, Windows 10

fabiankaegy:
@acosmin will get right on it. Thank you so much for all the work you have done :)

fabiankaegy:
@acosmin does it just not work at all or not smoothly scroll to the top? And are there any errors in the console?

acosmin:
`createEvent()` might also need to be prefixed

acosmin:
> @acosmin does it just not work at all or not smoothly scroll to the top? And are there any errors in the console?

No console errors, it jumps directly to header, no animations. It acts this way because the click event happens on the `span` tags, not the wrapping `a`. `pointer-events: none` solves that...

fabiankaegy:
@acosmin awesome! Did not know that worked 💯

acosmin:
works fine now 🎉

ianbelanger79:
Hmmm, it isn't working at all for me now. `Uncaught ReferenceError: createEvent is not defined
    at HTMLButtonElement.<anonymous> (index.js?ver=0.1:525)`

acosmin:
He forgot the `twentytwenty.` in front, it should work with that in

fabiankaegy:
🤦‍♂ I'm sorry  that was stupid

ianbelanger79:
No worries, that's why we test

fabiankaegy:
@ianbelanger79 @acosmin should be fixed :)

ianbelanger79:
Testing now

Issue #172, opened 2019-09-10
Remove deprecated live() from JS


mukeshpanchal27:
Google Chrome shows below error when we open the frontend.

jQuery.fn.live() is deprecated

andersnoren:
There's an existing issue and PR (#173) for removing jQuery entirely, so I'm closing this one.

PR #173, opened 2019-09-10
Remove deprecated live() from JS

mukeshpanchal27:
We have to use on method instead of live to fix the issue.

Fix: #172


andersnoren:
There's an existing issue and PR (#173) for removing jQuery entirely, so I'm closing this one. But thanks for submitting!

Issue #176, opened 2019-09-10
Don't escape translations when not necessary.


juanfra:
Just like core, default themes don't escape translations when not necessary as those are trusted.

See these links for background.

- https://core.trac.wordpress.org/ticket/30724 
- https://github.com/WordPress/twentynineteen/pull/285#issuecomment-432855498

ernilambar:
Although default theme does not follow this, I still think it is good to escape translation strings in all places, not only in HTML attributes.

mafzal9:
@nielslange please assign this issue to me.

nielslange:
> @nielslange please assign this issue to me.

Also, feel free just to say that you're working on it, so others know and won't interfere. 😀

mafzal9:
from now onward that's my issue, :) and i will try my best to contribute very well in WordPress community. :)

mafzal9:
@juanfra @ernilambar still i am confused did i need to remove escape translation from theme or add the escape translation in remaining tags  ?

juanfra:
Hello guys,

I hope you're all doing well. These issue was addressed in: https://github.com/WordPress/twentytwenty/pull/177 and https://github.com/WordPress/twentytwenty/pull/441

This one should be closed.

Best,
J.

mafzal9:
Thanks @juanfra for immediate response.

PR #177, opened 2019-09-10
Remove escape in translations.

juanfra:
Just like core, default themes don't escape translations when not necessary as those are trusted.

See these links for background.

https://core.trac.wordpress.org/ticket/30724
WordPress/twentynineteen#285 (comment)

*Fixes:* https://github.com/WordPress/twentytwenty/issues/176

WP Username: juanfra


pattonwebz:
While core does not escape translations I am not so certain that is the correct thing due to the possibility of unexpected entities being present in languages used by the translator. I left a comment on that ticket linked stating such but lets hear what others think about it too.

juanfra:
Thanks for the follow-up! There has been a discussion around this topic in a core ticket, but it was for the built-in themes.

Here's the discussion, where @ocean90 explains why these shouldn't be escaped: https://core.trac.wordpress.org/ticket/30724#comment:10

I've been searching a bit to see if there's a more recent discussion and it seems that for twentynineteen, the decision was the same: https://github.com/WordPress/twentynineteen/pull/523

pattonwebz:
I have looked into this some more and previous themes do not escape this but they do translate it. I will pull down this PR and add in a comment to allow travis to pass with this PR.

pattonwebz:
I rolled this PR into #441 and it got commit without me adding a close tag. I am closing this one here as the code is already merged but props should still be granted to `@juanfra` | WP Username: `juanfra`. #399

Issue #178, opened 2019-09-10
WordPress shouldn't be translated.


juanfra:
WordPress shouldn't be a translatable string.

ianbelanger79:
It has been discussed and WordPress does need to be translatable, so I am closing this as invalid

juanfra:
Thanks! do you have a link to the discussion?

PR #179, opened 2019-09-10
`WordPress` shouldn't be translatable.

juanfra:
WordPress shouldn't be translatable.

**Fixes:** https://github.com/WordPress/twentytwenty/issues/178

**WP Username:** juanfra


xlmnxp:
@juanfra why not?

juanfra:
Because it is a trademark, and it haven't been translated in any other theme.

xlmnxp:
@juanfra no
we write `trademark` in Arabic 
`WordPress` > `ووردبريس`

juanfra:
Thanks for your response. It seems that's always like that in core themes at least. You can take a look:

https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentynineteen/footer.php#L28

https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentyseventeen/template-parts/footer/site-info.php#L21

https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentysixteen/footer.php#L63

tobifjellner:
Even if we don't translate WordPress, some languages still need to transliterate it into other type of scripts.

ianbelanger79:
WordPress does need to be translatable, so I am closing this PR. Thanks for your submission @juanfra

juanfra:
@ianbelanger79 so you're saying that all of the other core themes are wrong?

tobifjellner:
@juanfra There was a short discussion in Slack: https://wordpress.slack.com/archives/C02RP4VMP/p1568130355315700

juanfra:
Thank you @tobifjellner! I see your interaction with `@poena` where she asks you and you confirm it shouldn't be translatable. I'm assuming then, that all of the other core themes are wrong.

tobifjellner:
You can make core tickets for the previous bundled themes, where applicable. :)

juanfra:
Cool, thanks!

senadir:
we can also make the assumption that other theme didn't consider it

Issue #180, opened 2019-09-10
Use template tags for the post ID.


juanfra:
Use `get_the_ID()` instead of the post property.

carolinan:
Ok so there is one $post->ID that did not make the merge:
https://github.com/WordPress/twentytwenty/blob/master/template-parts/entry-header.php#L52

juanfra:
Hey @carolinan! Thanks, added a new PR.

PR #181, opened 2019-09-10
Use template tags for the post ID.

juanfra:
Use get_the_ID() instead of the post property.

Fixes: https://github.com/WordPress/twentytwenty/issues/180
WP Username: juanfra


mukeshpanchal27:
Related: #146

PR #192, opened 2019-09-10
Fix all WPCS notices in the header.php file

pattonwebz:
This PR fixes the WPCS notices that were present in the header.php file

### DIFF
--- a/header.php
+++ b/header.php
@@ -1,3 +1,15 @@
+<?php
+/**
+ * Header file for the Twenty Twenty WordPress default theme.
+ *
+ * @link https://developer.wordpress.org/themes/basics/theme-functions/
+ *
+ * @package WordPress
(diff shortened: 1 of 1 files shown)


pattonwebz:
Thoughts on the file header comment I added here are welcomed.

header.php
```diff
@@ -1,3 +1,15 @@
+<?php
+/**
+ * Header file for the Twenty Twenty WordPress default theme.
+ *
+ * @link https://developer.wordpress.org/themes/basics/theme-functions/
```
ianbelanger79: I think that we may want to use https://developer.wordpress.org/themes/basics/template-files/#template-partials here

pattonwebz: Yup good catch, will sort this in a moment.

PR #193, opened 2019-09-10
Fix all WPCS notices in the footer.php file

ianbelanger79:
This is a first run, let me know if there is anything not correct


pattonwebz:
Approved here because I do not want a delay waiting on a decision to hold back going through and resolving as many of the issues as we can today. Merging what is possible now eases merge conflicts on PRs at later point.

PR #198, opened 2019-09-10
Fix all WPCS notices in the pagination.php file

pattonwebz:
Resolves some PHPCS notices in this file and adds contextual translation to pagination strings.

### DIFF
--- a/pagination.php
+++ b/pagination.php
@@ -1,4 +1,13 @@
 <?php
+/**
+ * A template partial to output pagination for the Twenty Twenty default theme.
+ *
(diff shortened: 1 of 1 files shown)


ernilambar:
May be we omit full stop `.` in file header, I guess. There seems other few files also have this same issue. Eg https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentynineteen/functions.php#L3

ntwb:
@ernilambar The "Summary" should *not* have a period, the "Description" should.

See https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers

PR #203, opened 2019-09-10
Use Standard Handles for scripts/styles

emiluzelac:
To avoid double loading, i.e. Plugins, third-party scripts, and styles should not be prefixed, here's some [background](https://github.com/grappler/wp-standard-handles). GH/WPORG username: emiluzelac


emiluzelac:
@pattonwebz thanks! And once again, I failed to add my usernames, so sorry! GB/WPORG emiluzelac :)

Issue #73, opened 2019-09-07
Use a `nav` element for all the menus and add labels


afercia:
All the UI sections meant for navigation should be wrapped in a `<nav>` element for better semantics and to allow assistive technology users to quickly jump through ARIA landmarks (`<nav>` elements are natively mapped to ARIA landmarks).

Currently all the three provided menus don't use a `<nav>` element:
- Main Menu
- Footer Menu 
- Social Menu 

I do realise the Main Menu is a special case, as it's hidden by default but it would be great to find a way to let assistive technologies understand there's a navigation menu at that location.

Also, when there's more than one `<nav>` element in a page, screen reader users need to be able to distinguish them. In absence of an accessible name, the `<nav>` elements are announced just as "navigation":

Screenshot with Safari + VoiceOver in a page where both the posts navigation and the pagination navigation are printed out:

![Screenshot 2019-09-07 at 14 42 01](https://user-images.githubusercontent.com/1682452/64475095-44b04d00-d17e-11e9-95e5-c75124c3f1f3.jpg)

Labeling is as simple as adding an `aria-label` attribute on the `<nav>` element, with a meaningful text. Please avoid the word "navigation": that's already announced as the role of the `<nav>`.

Examples:
- `<nav aria-label="Main menu" ...`
- `<nav aria-label="Secondary menu" ...`
- `<nav aria-label="Social links" ...`
- `<nav aria-label="Pagination" ...` (better wording welcome)

carolinan:
This is hard without being too repetitive. Perhaps:
More content (archives)
Pagination (in single post and page content)
Next and previous posts (single posts)
More comments ( comment pagination in single posts and pages)

afercia:
As a user, I'd need just a quick reference to understand _what_ a specific navigation is about. The word "navigation" is announced by the screen reader so even a very short label would work.

For example:
`Posts` and then the screen reader add `navigation`
`Comments` ...

Examples from core admin:

<img width="1159" alt="Screenshot 2019-09-07 at 18 15 42" src="https://user-images.githubusercontent.com/1682452/64477438-cf537500-d19b-11e9-833d-a3c9f8231e45.png">

carolinan:
I apologize, a simple test would of course have showed me this :)

carolinan:
What remains are the menus in the header / mobile menu.

pattonwebz:
[...]

PR #209, opened 2019-09-11
Add nav and aria labels to menus

carolinan:
Add nav and aria labels to menus
Partial fix for https://github.com/WordPress/twentytwenty/issues/73


carolinan:
This is only a partial fix since the main menus might still change.

Issue #210, opened 2019-09-11
Disable Search Button only applies to desktop


carolinan:
When the Disable Search Button option is enabled,
it only hides the search on desktop, not on smaller screens.

carolinan:
This has been solved and merged, closing.

nielslange:
> Hi, I'm having trouble finding where the option to Disable Search Button is. :(

@andrewtvuong: You can enable/disable the search button via:

`WP Admin —> Appearance —> Customise —> Theme Options`

as seen in 

![2DA9C363-D883-4610-9F3D-C6D10889F6ED](https://user-images.githubusercontent.com/3323310/69131213-8a758400-0ae4-11ea-9a91-a8709c890c7b.png)

andrewtvuong:
@nielslange thanks, I don't know why I kept missing it. New to CSS but went ahead and nuked the style.css already should anyone become interested in the future using

```
/* Search Toggle ----------------------------- */

.search-toggle {
	position: absolute;
	bottom: 0;
	left: 0;
	top: 0;
	display: none !important;
}

.search-toggle .toggle-icon,
.search-toggle svg {
	height: 2.5rem;
	max-width: 2.3rem;
	width: 2.3rem;
	display: none !important;
}
```

nielslange:
@andrewtvuong While this is also a solution, hiding the search icon via themes options is the better choice as it loads less code. In your solution, which is not bad, by the way, both the HTML code of the search icon and the CSS code to hide the search icon are transferred from the server to the client. While it's not much data that you avoid sending, every byte counts when it comes to performance. 

Hiding the search icon via theme options, simply removes the HTML part for the search icon from the site, while hiding the search icon via CSS will require to load the following code, even though nothing is visible to the visitor:

**HTML part**

```
<div class="header-toggles hide-no-js">
  <div class="toggle-wrapper search-toggle-wrapper">
    <button class="toggle search-toggle desktop-search-toggle" data-toggle-target=".search-modal" data-toggle-body-class="showing-search-modal" data-set-focus=".search-modal .search-field" aria-expanded="false">
      <span class="toggle-inner">
[...]

PR #213, opened 2019-09-11
Disable search button in mobile view

acosmin:
Fix for #210

### DIFF
--- a/header.php
+++ b/header.php
@@ -42,7 +42,7 @@
 					// Check whether the header search is deactivated in the customizer.
 					$disable_header_search = get_theme_mod( 'twentytwenty_disable_header_search', false );
 
-					if( get_theme_mod( $disable_header_search ) ) {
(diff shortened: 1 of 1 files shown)


carolinan:
replace the : and the endif with braces

acosmin:
@carolinan I thought of doing that :) but the entire file uses that style... should I really do it?

acosmin:
The file needs an entire rewrite

acosmin:
Done :)

acosmin:
Wait, I made a little mistake

carolinan:
I believe the next big change to this file will be if the menus are changed. So a full rewrite should probably be made then.

acosmin:
Anyway, I think I got it right this time. It works...

acosmin:
@carolinan I think this one can be merged...

acosmin:
Redoing this...

PR #217, opened 2019-09-11
Fix #205: Replace https://www.wordpress.org to https://wordpress…

### LINKED ISSUE #205: WordPress URL in footer
We generally omit `www.` in WordPress.org URL.

URL - `https://www.wordpress.org`

This could be simply `https://wordpress.org/`

Also note, trailing slash.

burhandodhy:


ernilambar:
@burhandodhy Can you please update PR so that only line 24 (URL) only is changed? There seems to other changes also.

burhandodhy:
@ernilambar only the trailing spaces are removed on the other lines. There is no other change

ntwb:
@ernilambar If a PR contains extra lines where trailing whitespace is removed that is fine, its not that easy to change ones Editor/IDE to leave these in tact, and as the WP CS require trailing whitespace be removed its easier to just do it when it occurs.

PR #227, opened 2019-09-11
UPDATE #212: Adjust search button option

### LINKED ISSUE #212: Discussion: Rephrase the Disable Search Button option
This is the current text for the search button option in the customizer

----

Settings for the site header.

- [ ] Disable Search Button

_Check to disable the search button in the header._

----

I think it would make more sense for this option to be reversed;
for the search to be enabled by default, with the box checked,
and disabled when unchecked. 

I don't think the word "Button" is needed.

The text could be rephrased to something like:

----

Settings for the site header.

- [x] Enable search

_Uncheck to disable the search in the header._

----

Currently, the Disable Search Button option is in the site header panel.
**-I keep clicking this panel expecting to add a header image.**
And this is is despite knowing very well that the header image panel has a different name :)

nielslange:
Could fix #212 (if the wording and the behaviour should be changed). See also https://github.com/WordPress/twentytwenty/pull/225#ref-pullrequest-492326990

### DIFF
--- a/classes/class-twentytwenty-customize.php
+++ b/classes/class-twentytwenty-customize.php
@@ -116,25 +116,25 @@ if ( ! class_exists( 'TwentyTwenty_Customize' ) ) {
 				)
 			);
 
-			/* Disable Header Search --------- */
+			/* Enable Header Search --------- */
 
 			$wp_customize->add_setting(
-				'twentytwenty_disable_header_search',
+				'twentytwenty_enable_header_search',
 				array(
 					'capability'        => 'edit_theme_options',
-					'default'           => false,
+					'default'           => true,
 					'sanitize_callback' => 'twentytwenty_sanitize_checkbox',
 				)
 			);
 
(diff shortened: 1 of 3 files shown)


carolinan:
The PR works well but function twentytwenty_body_classes has now been moved from functions.php to https://github.com/WordPress/twentytwenty/blob/master/inc/template-tags.php#L404

nielslange:
@carolinan How's `twentytwenty_body_classes` related to this PR?

carolinan:
> @carolinan How's `twentytwenty_body_classes` related to this PR?
It's the name if the function where the different body class is added depending on the theme mod, 

See https://github.com/WordPress/twentytwenty/pull/227/files#diff-78cd5aa3783a74555c9938a2a81d01c6R216

nielslange:
Thanks, @carolinan! Now I see what you mean and I'll update this PR accordingly.

nielslange:
Closed in favour of #252

Issue #235, opened 2019-09-12
Implement the Image Overlay Text Color option


carolinan:
This text color option for the cover template has not been implemented yet,
(https://github.com/WordPress/twentytwenty/blob/master/classes/class-twentytwenty-customize.php#L232)

The option is _twentytwenty_cover_template_overlay_text_color,_ but please pay close attention since the prefix here _will be removed._ See: https://github.com/WordPress/twentytwenty/issues/137

It then needs to be added to twentytwenty_get_customizer_css.
https://github.com/WordPress/twentytwenty/blob/master/functions.php#L1018

The color also needs to be applied on the post meta, see https://github.com/WordPress/twentytwenty/issues/218

carolinan:
This has been merged with https://github.com/WordPress/twentytwenty/pull/245 , closing.

PR #245, opened 2019-09-13
Implemented the Image Overlay Text Color option

acosmin:
Fix for #235

Prefixes should be handled in another PR

Also, I changed the defaults to lowercase to match what the Customizer color picker saves.


carolinan:
twentytwenty_get_customizer_css has moved to inc/custom-css.php

acosmin:
I'm working on it right now :)

acosmin:
@carolinan done, did you guys remove the post meta and excerpt from the cover overlay?

carolinan:
Hmm no, that was not intentional, posts are meant to show both, pages are only meant to show the arrow.

acosmin:
If you want I can redo them in this PR

carolinan:
Ok, I trust that you see which code that needs to be moved from entry-header to content-cover :)

acosmin:
@carolinan it should work now :)

carolinan:
I can't find a mockup that shows the excerpt or meta, but looking back,
the original page-header.php only showed the arrow for pages, not for posts.
Posts showed excerpt and meta.

carolinan:
entry-header.php is not used for the cover template, so ideally the code for the template should be removed from this file.

acosmin:
I think the arrow was available for posts as well, from what I remember...

andersnoren:
The arrow was only displayed for pages with the cover template, originally. It gets a bit messy with posts that have both a title, excerpt, meta items, and the arrow.

acosmin:
Rearranged them... :)

Issue #246, opened 2019-09-13
Update the site header and menus to match new structure


andersnoren:
This issue details the implementation of the new header structure, based on the discussions in the following issues: #164  #74. Thanks to everyone who has contributed to the discussion. I’ve posted a summary of the reasoning behind the final structure in this issue: #164. It’s on the lengthy side, so I won’t repeat all of it here.

The new structure will require changes to the menu locations, the output of the menus and some surrounding elements, and the styling. I’ve split them up into three headings below.

~~**NOTE:** The new folder structure should be implemented before we start work on this. See #244~~

# Menu locations

The following menu locations will be used:

- `primary` Desktop Horizontal Menu
- `expanded` Desktop Expanded Menu
- `mobile` Mobile Menu
- `footer` Footer Menu
- `social` Social Menu

The ID’s have been selected to help WordPress map existing menus to their corresponding locations in Twenty Twenty when users activate the theme. More info on that in #164.

# Output changes

## header.php
- The `wp_nav_menu()` call (and corresponding `has_nav_menu()` conditional) for the primary menu (previously shortcuts menu) needs to be updated with the new `primary` menu location name.
- If the `primary` menu is the primary menu, we should include the `wp_list_pages()` fallback here instead of in the `modal-menu.php`.
- The class names and selectors for the shortcuts menu should probably be changed to be called `primary` instead of `shortcuts`, for consistency.

## template-parts/modal-menu.php
The styling for the `mobile` and `expanded` menu is shared, so they should have a shared class that can be used for that targeting. It’s currently using `.main-menu` as a top level selector for the menu styles.

If both the `mobile` and `expanded` locations are set, and they are set to two different menus, both should be output with two different `wp_nav_menu` calls in the modal menu. The `mobile` one should be displayed up to `1000px` width, and the `expanded` one should be displayed from `1000px` and up.

If both menu locations are set, but to the same menu, only one element should be output, to save on overall page size.

If the `expanded` menu isn’t set, it shouldn’t be output, and the menu toggle should be hidden on `1000px` and up.

If the mobile menu isn’t set, we should first check if a menu is set to the  `primary` location, and if it is, use that menu instead. If not, check if `expanded` location is set, and if it is, use that menu. If neither are set, use a `wp_list_pages()` fallback.

The `wp_nav_menu()` call for the social menu needs to be updated with the new `social` menu location.

[...]

PR #247, opened 2019-09-13
Update Menu names

### LINKED ISSUE #73: Use a `nav` element for all the menus and add labels

### LINKED ISSUE #165: declare menus in order of importance
On the admin Menus page and in the Customizer, the menus are presented in the order of registration. It would be nice to have the primary or main menu shown first and the footer menu at the bottom.

https://github.com/WordPress/twentytwenty/blob/7a0f2da2798bf052f5f5aa5a2ef4d68f83b30095/functions.php#L172-L177

carolinan:
Partial fix for #246
Updates functions.php, template-tags.php, footer.php, header.php. wp_list_pages fallback, change CSS class name from shortcuts to primary.
**Only minor changes to modal-menu.php.**

Fixes #73 (nav aria labels)
Fixes #165


andersnoren:
Looks good to me. 

@carolinan Do you want to add to this PR with the conditional outputs and fallbacks in `modal-menu.php`, or do you think that should be done with a separate PR?

carolinan:
I did not fully understand how those conditional outputs and fall backs would work from only reading your post, so I left it out.

PR #248, opened 2019-09-13
Use `get_the_ID()`

juanfra:
Use get_the_ID() instead of the post property.

Fixes: #180
WP Username: juanfra


carolinan:
Thank you!

PR #249, opened 2019-09-13
Auto-calculate background-color luminance (#a11y)

aristath:
See #3 

This is a PHP implementation that changes a body class depending on the luminance of the selected background-color.
Currently it adds `background-color-light` and `background-color-dark` classes depending on the relative luminance of the color.
This PR only imcludes a PHP implementation because from what I've seen so far the control has `'transport' => 'refresh'` instead of `'transport' => 'postMessage'`.
Once the transport for this control changes I can push another PR to handle the JS side of things in the customizer.

The same implementation can be used to change the editor to dark-mode, it's basically just 2 lines to get the luminance and figure out if the color is light or dark.

The calculations use the formula for relative luminance from https://www.w3.org/WAI/GL/wiki/Relative_luminance instead of a generic brightness implementation. This way we're getting true perceivable luminance instead of relative to the hue etc.

### DIFF
--- a/classes/class-twentytwenty-color.php
+++ b/classes/class-twentytwenty-color.php
@@ -16,17 +16,17 @@ if ( ! class_exists( 'TwentyTwenty_Color' ) ) {
 		/**
 		 * The HEX color.
 		 *
-		 * @access private
-		 * @since 1.0
+		 * @access protected
+		 * @since 1.0.0
 		 * @var string
 		 */
-		private $hex = '';
+		protected $hex = '';
 
 		/**
 		 * Sets the $hex property of this object.
 		 *
 		 * @access public
-		 * @since 1.0
+		 * @since 1.0.0
 		 * @param string $hex The HEX color.
 		 * @return void
 		 */
@@ -38,7 +38,7 @@ if ( ! class_exists( 'TwentyTwenty_Color' ) ) {
 		 * Sanitizes a HEX color.
 		 *
 		 * @access public
-		 * @since 1.0
+		 * @since 1.0.0
 		 * @param string $color The color to sanitize.
 		 * @return string
 		 */
@@ -59,7 +59,7 @@ if ( ! class_exists( 'TwentyTwenty_Color' ) ) {
 		 * Get RGB as an array from $this->hex.
 		 *
 		 * @access public
-		 * @since 1.0
+		 * @since 1.0.0
 		 * @return array
 		 */
 		public function get_rgb() {
@@ -77,7 +77,7 @@ if ( ! class_exists( 'TwentyTwenty_Color' ) ) {
 		 * The sRGB space is used to properly calculate relative luminances.
 		 *
 		 * @access public
-		 * @since 1.0
+		 * @since 1.0.0
 		 * @return array
 		 */
 		public function get_srgb() {
--- a/classes/class-twentytwenty-color.php
+++ b/classes/class-twentytwenty-color.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * Color calculations.
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty
+ * @since 1.0.0
+ */
+
+if ( ! class_exists( 'TwentyTwenty_Color' ) ) {
+	/**
+	 * Color Calculations.
+	 */
+	class TwentyTwenty_Color {
+
+		/**
+		 * The HEX color.
+		 *
+		 * @access private
+		 * @since 1.0
+		 * @var string
+		 */
+		private $hex = '';
+
+		/**
+		 * Sets the $hex property of this object.
+		 *
+		 * @access public
+		 * @since 1.0
+		 * @param string $hex The HEX color.
+		 * @return void
+		 */
+		public function set_hex( $hex ) {
+			$this->hex = $this->sanitize_hex( $hex );
+		}
+
+		/**
+		 * Sanitizes a HEX color.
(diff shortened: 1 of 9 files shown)


aristath:
Sample CSS:
To test this implementation you can just add this in the customizer's custom-css field:

```css
.background-color-dark {
	color: #fff;
}
```
and then play with the background colors :)

carolinan:
> This PR only imcludes a PHP implementation because from what I've seen so far the control has `'transport' => 'refresh'` instead of `'transport' => 'postMessage'`.
> Once the transport for this control changes I can push another PR to handle the JS side of things in the customizer.

The transport has been updated.

aristath:
Hmmm I see there's no `customizer.js` file.
This should probably be added & enqueued first.
I _could_ add it of course, and write the scripts required to make postMessage work for the background color etc, but I assume that should be done in another ticket with another PR and not here?

classes/class-twentytwenty-color.php
```diff
@@ -0,0 +1,103 @@
...
+	class TwentyTwenty_Color {
+
+		/**
+		 * The HEX color.
+		 *
+		 * @access private
+		 * @since 1.0
+		 * @var string
+		 */
+		private $hex = '';
```
pattonwebz: Wht do you think about making this `protected` instead?

classes/class-twentytwenty-color.php
```diff
@@ -0,0 +1,103 @@
...
+	/**
+	 * Color Calculations.
+	 */
+	class TwentyTwenty_Color {
+
+		/**
+		 * The HEX color.
+		 *
+		 * @access private
+		 * @since 1.0
```
pattonwebz: When since strings are specified could we use 3 digit references always? - `1.0.0`

classes/class-twentytwenty-color.php
```diff
@@ -0,0 +1,103 @@
...
+				'r' => $rgb['r'] / 255,
+				'g' => $rgb['g'] / 255,
+				'b' => $rgb['b'] / 255,
+			];
+		}
+
+		/**
+		 * Gets the relative luminance of a color.
+		 *
+		 * @see https://www.w3.org/WAI/GL/wiki/Relative_luminance
```
pattonwebz: I assume this is the best link to reference this formulae?

pattonwebz:
Hey @aristath thanks so much for this. I didn't expect you'd be able to put something workable like this together quite so fast :)

Looks good to me mostly, left a minor change request and a couple question in the review.

classes/class-twentytwenty-color.php
aristath: I see no problem with making it `protected`, I'm just used to making `private` anything that should not be directly accessed ;)

I'll push a commit shortly

classes/class-twentytwenty-color.php
aristath: There's some wikipedia links too, but since this is the "official" w3.org site it seemed more appropriate

aristath:
Great. I pushed the requested changes already.
As soon as there's a `customizer.js` file with a working implementation for `background_color` I can push the JS implementation too. :)

classes/class-twentytwenty-color.php
pattonwebz: Yeah this seems most suitable of them all then :)

celloexpressions:
This is a good start at implementing contrast checking functionality. I think we need to take a broader look at the various places that colors are customizable and the different potential combinations of colors. Then, we can come up with a comprehensive approach to making colors customizable while automatically maintaining contrast. We can finish implementing the calculations library in the meantime.

Here are a few specific comments on the current implementation in this PR:
1. Can we call the color class and file "Color Calculations" instead of just "Color"? This avoids potential confusion that this file may contain patterns for color usage, as opposed to calculation helpers. I typically use a `color-calculations.php` file for the calculations and a `color-patterns.php` file for the CSS patterns (we already have a separate file for CSS, here).
2. I recommend including a contrast ratio function, that takes two hex colors as inputs and returns the contrast ratio as defined by WCAG. This allows the usage of color calculations to directly relate to the typical contrast ratio guidelines, rather than relying on relative luminance in specific color checks. Example: `if ( 4.5 < twentytwenty_contrast_ratio( $background_color, '#000' ) ) { $dark_mode = true; }`
3. It is also useful to include a function that can lighten or darken a user-selected color as required to achieve a required contrast ratio. We would likely use this for the accent color to maintain contrast with the background color (since it wouldn't be able to be a direct black/white switch like the text color). This functionality is generally required if any of the user-customizable colors need to contrast with both a light and a dark color.
4. I suggest looking at the color calculation library developed with the Fourteen Colors plugin (originally part of the Twenty Fourteen theme, and moved to a plugin due to limited development time) for a starting point for 2 and 3. These functions have worked well without updates for six years; some modernization would be good, but there is also deep usage in several themes. See https://plugins.trac.wordpress.org/browser/fourteen-colors/trunk/color-calculations.php
5. Let's address instant live preview of customized colors separately from the initial implementation of automated color contrast checks. Both of these topics require a comprehensive look at all of the customizable colors in the theme and can be implemented independently once the color pattern strategy is refined.
6. All of the color calculations can (and should) remain in PHP only. The live preview JS can swap out direct color changes without rebuilding the CSS, and selective refresh can be used to periodically update calculated color changes from the PHP code. __We should not duplicate color contrast calculations or CSS color patterns in both PHP and JS__.

PR #251, opened 2019-09-13
UPDATE #250: Update structure

### LINKED ISSUE #250: Update structure
Things to do:

- [x] Move `template-cover.php` into `templates` folder
- [x] Move `template-full-width.php` into `templates` folder
- [x] Move `twentytwenty-editor-style-block-editor.css` into `assets/css/`
- [x] Move `twentytwenty-editor-style-classic-editor.css` into `assets/css/`
- [x] Rename `twentytwenty-editor-style-block-editor.css` to `editor-style-block.css`
- [x] Rename `twentytwenty-editor-style-classic-editor.css` to `editor-style-classic.css`

I'm currently working on this issue and will create a PR later today.

nielslange:
Fixes #250


inc/template-tags.php
```diff
@@ -448,12 +448,12 @@ function twentytwenty_body_classes( $classes ) {
 		}
 
 		// Check whether the current page should have an overlay header.
-		if ( is_page_template( array( 'template-cover.php' ) ) ) {
+		if ( is_page_template( array( 'templates/template-cover.php' ) ) ) {
```
joyously: Did you try this? Does that function return a path or just a name?

inc/template-tags.php
```diff
@@ -448,12 +448,12 @@ function twentytwenty_body_classes( $classes ) {
...
 			$classes[] = 'overlay-header';
 		}
 
 		// Check whether the current page has full-width content.
-		if ( is_page_template( array( 'template-full-width.php' ) ) ) {
+		if ( is_page_template( array( 'templates/template-full-width.php' ) ) ) {
```
joyously: Can't we just use the existing classes added for page templates?

inc/template-tags.php
nielslange: It simply returns true, if the page was found, or false, if the page wasn't found: https://developer.wordpress.org/reference/functions/is_page_template/#return

inc/template-tags.php
nielslange: Do you mean moving the files into the `classes` folder instead of into a `templates` folder?

joyously: No. I mean that WP already puts the page template slug as a class on the body. Why can't we just use that class instead of adding more?

inc/template-tags.php
joyously: According to that code, it would never match because it's looking at the slug, not the path.

inc/template-tags.php
joyously: And do they really need to be body classes?

singular.php
```diff
@@ -21,7 +21,7 @@
 		while ( have_posts() ) {
 			the_post();
 
-			if ( is_page_template( array( 'template-cover.php' ) ) ) {
+			if ( is_page_template( array( 'templates/template-cover.php' ) ) ) {
 				get_template_part( 'template-parts/content-cover' );
```
joyously: The intent was for this to move to the actual template file instead of the template just calling singular. Keep the template-specific stuff in the template as much as possible.

inc/template-tags.php
nielslange: I believe @andersnoren added them initially so that folks can target these pages via CSS later on. But you're correct. When selecting the template `Full Width Template` the classes `page-template-template-full-width` and `page-template-template-full-width-php` are already added to the body class. I'll update the PR and remove them for now.

singular.php
nielslange: I'm afraid, I don't understand what you mean by that. Would you mind explaining it to me?

joyously: The template-cover file should contain a loop that uses the content-cover template part.
The singular.php file should contain a loop that uses the content template part.
There should not be a check for the page template in either one.

carolinan: I agree with Joy, this solution that I added when the file(s) were first split is not ideal.

template-parts/entry-header.php
```diff
@@ -31,8 +31,27 @@
 			<?php
 		}
 
-		// Default to displaying the post meta.
-		twentytwenty_the_post_meta( get_the_ID(), 'single-top' );
+		// On pages with the cover template, display a "To the content" link.
```
carolinan: The cover template does not use entry-header.php, the related code is now in https://github.com/WordPress/twentytwenty/blob/master/template-parts/content-cover.php#L56

singular.php
nielslange: @joyously & @carolinan I've adjusted the files as suggested by you in https://github.com/WordPress/twentytwenty/pull/251/commits/cc842420774c7851ff9e73cd06391775f0f23e96

template-parts/entry-header.php
nielslange: @carolinan I resolved that conflict in https://github.com/WordPress/twentytwenty/pull/251/commits/b5df3b262fd82ecfe1a8670fa115852fc97b96c9

PR #252, opened 2019-09-13
UPDATE #212-2: Adjust search button option

### LINKED ISSUE #212: Discussion: Rephrase the Disable Search Button option

nielslange:
Fixes #212

### DIFF
--- a/header.php
+++ b/header.php
@@ -115,10 +115,10 @@
 
 						<?php
 
-						// Check whether the header search is deactivated in the customizer.
-						$disable_header_search = get_theme_mod( 'twentytwenty_disable_header_search', false );
+						// Check whether the header search is activated in the customizer.
+						$enable_header_search = get_theme_mod( 'twentytwenty_enable_header_search', false );
 
-						if ( ! $disable_header_search ) :
+						if ( true === $enable_header_search ) :
 							?>
 
 							<div class="toggle-wrapper search-toggle-wrapper">
@@ -139,8 +139,8 @@
 			</div><!-- .header-inner -->
 
 			<?php
-			// Output the search modal (if it isn't deactivated in the customizer).
-			if ( ! $disable_header_search ) {
+			// Output the search modal (if it is activated in the customizer).
+			if ( true === $enable_header_search ) {
 				get_template_part( 'template-parts/modal-search' );
 			}
 			?>
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -457,9 +457,9 @@ if ( ! function_exists( 'twentytwenty_body_classes' ) ) {
 			$classes[] = 'has-full-width-content';
 		}
 
-		// Check for disabled search.
-		if ( get_theme_mod( 'twentytwenty_disable_header_search', false ) ) {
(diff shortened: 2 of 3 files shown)


inc/template-tags.php
```diff
@@ -457,9 +457,9 @@ function twentytwenty_body_classes( $classes ) {
 			$classes[] = 'has-full-width-content';
 		}
 
-		// Check for disabled search.
-		if ( get_theme_mod( 'twentytwenty_disable_header_search', false ) ) {
-			$classes[] = 'disable-search-modal';
+		// Check for enabled search.
+		if ( true === get_theme_mod( 'twentytwenty_enable_header_search' ) ) {
+			$classes[] = 'enable-search-modal';
```
joyously: Does this class affect any CSS?

nielslange: @ joyously No, currently there is no CSS class `.disable-search-modal`. I believe @andersnoren added that body class to enable users to add conditional CSS, when needed, later on.

joyously: OK, well, I just searched the files for 'search-modal' and the JS uses `search-modal`, which the CSS uses also. The header and the modal-search file both have the `search-modal` class, but also have data attributes 
`data-toggle-target=".search-modal"`
`data-toggle-body-class="showing-search-modal"`
`data-set-focus=".search-modal .search-field"`

Seems like the body class should match. (and 'showing' makes more sense than 'enable')

header.php
```diff
@@ -115,10 +115,10 @@
 
 						<?php
 
-						// Check whether the header search is deactivated in the customizer.
-						$disable_header_search = get_theme_mod( 'twentytwenty_disable_header_search', false );
+						// Check whether the header search is activated in the customizer.
+						$enable_header_search = get_theme_mod( 'twentytwenty_enable_header_search', false );
```
andersnoren: This should be updated to have `true` as default value, to match the `add_control()`.

carolinan:
Sorry about this mess I made :) . -Please also hide the mobile search form, starting on line 40 in header.php.

header.php
nielslange: Nice catch, @andersnoren! I've fixed this issue in https://github.com/WordPress/twentytwenty/pull/264

nielslange:
> Please also hide the mobile search form toggle, starting on line 40 in header.php.

@carolinan With https://github.com/WordPress/twentytwenty/pull/264 the visibility of the mobile search icon can be toggled. Feel free to merge this PR again, if you believe it's in a good shape now.

Issue #128, opened 2019-09-08
Use core functionality for custom logo


joyously:
Themes should use core functionality if it exists, so these two functions for custom logo should be removed. Adjusting the generated HTML should be accomplished with a filter.

https://github.com/WordPress/twentytwenty/blob/37d7eddf5817f15b67e9707841c889a78d2b29de/functions.php#L277-L286

Core functions: https://developer.wordpress.org/?s=custom_logo
https://developer.wordpress.org/reference/functions/add_theme_support/#custom-logo

byalextran:
is this as straightforward as it seems? here's what i'm thinking would change: 

1. replace existing custom logo code from `functions.php` with `add_theme_support( 'custom-logo' )`. 
2. update `header.php` to use core functions for custom logos. 

if it is that simple, i'd like to take a stab at a PR.

Issue #129, opened 2019-09-08
remove checks for hook functions


joyously:
Most of the functions in functions.php are either filters or actions. These should not have an if statement for a function definition, making them pluggable. If other code (child theme or plugin) wants to change one of these hooks, they can remove the theme hook, and write a hook of their own with their own prefix. 

https://github.com/WordPress/twentytwenty/blob/37d7eddf5817f15b67e9707841c889a78d2b29de/functions.php#L12

ianbelanger79:
We will be removing the function_exists checks

PR #256, opened 2019-09-14
Used core functions for custom logo and updated styling.

byalextran:
Fixes https://github.com/WordPress/twentytwenty/issues/128. 

My first contribution to a WordPress project. I hope everything's up to par. 🤞

**Note**: I removed the `function_exists` call based on https://github.com/WordPress/twentytwenty/issues/129. 

WordPress Username: @byalextran


joyously:
Hi Alex!
I think this should be handled differently. I'm not sure what `twentytwenty_retina_logo` is, but what if that is checked at the time of setup for the custom logo:
https://github.com/WordPress/twentytwenty/blob/f03340e728b4fe052d7cbfa425bbc6f1e715294c/functions.php#L68

This function is doing pretty much what the core function already does, but it handles multisite also.
https://developer.wordpress.org/reference/functions/get_custom_logo/

Also, when you submit a PR, the coding standards will be checked by Travis. You have to click the Details link and then the build that failed to get the error message for what to change.

byalextran:
@joyously thanks for walking me through this. 

i'm honestly not sure how moving the `twentytwenty_retina_logo` to `twentytwenty_theme_support()` would work. for example, would we need to store the width/height of the image somewhere to then reference later when outputting the logo? 

or could it be as simple as taking the default output of `get_custom_logo` and just modifying the width/height values according to the `twentytwenty_retina_logo` setting? so not building the HTML from scratch like `twentytwenty_get_custom_logo()` is doing currently. 

as for the build failing, this appears to be the error: 

```
Script @php ./vendor/squizlabs/php_codesniffer/bin/phpcs --runtime-set testVersion 5.6- handling the checkcs event returned with error code 2
The command "if [[ "$PHPCS" == 1 ]]; then composer checkcs; fi" exited with 2.
```

i have no clue what that means or how to fix it. any insight on it?

joyously:
I think if you adjust the width and height in the add_theme_support call, you can just call `the_custom_logo()` to output it, as usual.

For the build, scroll to just above the red part to see that there is one error and one warning. It gives the file name, then line number and detailed reason, sort of in a box.

byalextran:
ooh, i think i understand now. i attempted to do that with the following code: 

```php
		// Custom logo.
		$logo_id = get_theme_mod( 'custom_logo' );
		$logo_width  = 320;
		$logo_height = 240;

		if ( $logo_id ) {
			$logo = wp_get_attachment_image_src( $logo_id, 'full' );

			if ( $logo ) {
				// For clarity.
				$logo_width  = esc_attr( $logo[1] );
				$logo_height = esc_attr( $logo[2] );

				// If the retina logo setting is active, reduce the width/height by half.
				if ( get_theme_mod( 'twentytwenty_retina_logo', false ) ) {
					$logo_width  = floor( $logo_width / 2 );
					$logo_height = floor( $logo_height / 2 );
				}
			}
		}

		add_theme_support(
			'custom-logo',
			array(
				'height'      => $logo_height,
				'width'       => $logo_width,
				'flex-height' => true,
				'flex-width'  => true,
				'header-text' => array( 'site-title', 'site-description' ),
			)
		);
```

however, it appears regardless of the width/height set in the `add_theme_support` call (or if `flex-height/width` are set to `false`), the output of `get_custom_logo` is the actual width/height of the image. i'm wondering now if that behavior is what caused @andersnoren to use a theme-specific version of `get_custom_logo`? 

perhaps he could chime in here?

also, i see now the error i have to fix with the build. thanks for pointing me in the right direction.

andersnoren:
Hi @byalextran,

Thanks for taking a look at this! First, the recommended logo width and height in there now is from Chaplin, so we'll need to adjust them a bit. Currently, the logo in Twenty Twenty is never displayed bigger than with a 9rem max height, so the recommended values can be:

```
$logo_width  = 120;
$logo_height = 90;
```

For the `add_theme_support` call, I think we want to check if the user has activated the `twentytwenty_retina_logo` setting, and if they have, double the recommended values, letting the user know that we're expecting an image with that resolution. So:

```
// Custom logo.
$logo_width  = 120;
$logo_height = 90;

// If the retina setting is active, double the recommended width and height.
if ( get_theme_mod( 'twentytwenty_retina_logo', false ) ) {
	$logo_width  = floor( $logo_width * 2 );
	$logo_height = floor( $logo_height * 2 );
}

add_theme_support(
	'custom-logo',
	array(
		'height'      => $logo_height,
		'width'       => $logo_width,
		'flex-height' => true,
		'flex-width'  => true,
		'header-text' => array( 'site-title', 'site-description' ),
	)
);
```

In `twentytwenty_get_custom_logo()`, we'll want to divide the width and height of the logo being output by half, so same as the code you have currently. I think that should cover it. The user could still upload an image with a higher or lower resolution than the recommended values, but as long as the we divide the width and height attributes of the image by two during output, the result will always be "retina".

---

While you're making changes to the logo stuff, could you add the following logo specific CSS?

After the `.header-titles {` style at the `@media ( min-width: 1000px )` breakpoint, around line 3625, add the following:

```
.wp-custom-logo .header-titles {
	align-items: center;
}
```

This will set the site tagline to align to the middle of the custom logo (rather than the bottom, which it does when a site title is output).

At line 1197, in the `.site-logo img {` style, add the following:

```
margin: 0 auto;
```

...making sure that the logo image is centered on table screens.

We'll have to think a bit on how to handle the main menu top position when a logo is set, since the extra header height added by the logo results in some of the menu being obscured. I think we might need a JavaScript based solution that calculates the height of the site header and offsets the `.menu-modal` top position accordingly.

byalextran:
@andersnoren thanks for the additional insight. i think i can definitely do those things. 

one point of clarification though. 

i believe @joyously's original concern was that `twentytwenty_get_custom_logo` repeated much of the logic already contained within the core  `get_custom_logo` function. so one of the goals (from what i understood) was to move much of the custom logic into core functions. 

would there be any issue with removing most of the output logic within `twentytwenty_get_custom_logo` by making a call to `get_custom_logo` and just modifying the width/height based on the retina setting? that way the default output of `get_custom_logo` is retained (and the benefits associated with that, like multisite support) while still supporting retina logos. 

i did notice the custom function adds an additional css class. would that need to be added as well?

byalextran:
@andersnoren @joyously alrighty folks. swing number two. see if that approach works. basically, retained the default output of `get_custom_logo` and used `preg_replace` to swap out the width/height values if the retina setting is enabled.

byalextran:
also, as i was looking at the default output of `get_custom_logo`, i noticed the use of `srcset`. does that feature render the need for the retina option obsolete? 

https://make.wordpress.org/core/2015/11/10/responsive-images-in-wordpress-4-4/

andersnoren:
@byalextran Using `preg_replace` seems much cleaner – good call. 

Re: `srcset`, we'll still need to modify the `width` and `height` values of the image to make it max out at half of its uploaded size, so `srcset` isn't much help there. We also need the option to modify the recommended width and height when they upload their logo.

It looks like the `style.css` file in your commit contains a lot of whitespace changes, which conflicts with the main branch. Can you remove those changes from this PR? When you do, I can merge this into master.

byalextran:
@andersnoren done!

andersnoren:
Thanks @byalextran! Merged. I'll add you to the contributors file as well :)

Issue #257, opened 2019-09-14
`TwentyTwenty_Customize` Class methods shouldn't be prefixed


acosmin:
Example:
https://github.com/WordPress/twentytwenty/blob/058b5783f2b1ece2d0fdb2c46f2437cd92836a31/classes/class-twentytwenty-customize.php#L21

acosmin:
This is solved, closing

PR #258, opened 2019-09-14
`TwentyTwenty_Customize` Class Rework

acosmin:
+ made the sanitize functions part of the class

Fix for #257


acosmin:
Removed `type => theme_mod` and `settings => settingname` as they are not needed.

acosmin:
@carolinan I think this can be merged :) so we can all work with the same options and avoid conflicts.

PR #237, opened 2019-09-12
UPDATE #231: Show .alignleft and .alignright as block elements on small devices

nielslange:
Fixes #231 

<table>
<tr>
<td>Before (414px * 736px):
<br><br>

</td>
<td>After (414px * 736px):
<br><br>

</td>
</tr>
</table>

PR #236, opened 2019-09-12
UPDATE #231: Adjust .alignleft and .alignright on table and up

nielslange:
Fixes #231 

<table>
<tr>
<td>Before (1024px * 1366px):
<br><br>

</td>
<td>After (1024px * 1366px):
<br><br>

</td>
</tr>
</table>

<table>
<tr>
<td>Before (1920px * 1200px):
<br><br>

</td>
<td>After (1920px * 1200px):
<br><br>

</td>
</tr>
</table>

PR #259, opened 2019-09-14
Update block styles

### LINKED ISSUE #67: Form controls miss a focus style
Clear indication of focus is essential for keyboard navigation.

Currently, there's no indication of focus on input fields, textarea, select elements. The current CSS resets native focus style with `outline: none;` and doesn't provide an alternative style.

This affects all form controls displayed on the front end, for example the Search input field or the Comments form.

The caret isn't a sufficient indication of focus. Color alone isn't either. Focus needs to be indicated with a shape. There are two options:

- either remove `outline: none;` and rely on the browsers native focus style
- or provide an alternative focus style which _improves_ the native one

Screenshot of the comments form: tab through the form elements and notice only the checkbox has a focus style:

Screenshot 2019-09-07 at 13 55 13

Screenshot of the search form: no focus style:

Screenshot 2019-09-07 at 14 04 40

Of course, this applies also to buttons and any other interactive control but I see there's no `outline: none` there, though the focus style could be improved a bit.

### LINKED ISSUE #88: big tag, code tag, small tag - size - and quote tag italics
The `<big>` tag is not big.
The `<code>` tag is _too_ big.
The `<small>` tag is not small.
The `<q>` tag is italic. (but it can have nested `<em>`)

Legacy content can have deprecated HTML, but they should still be styled.
This is all related to #6.

big-and-code-size

quote-and-small

### LINKED ISSUE #254: Block: Separator
In the design, there are three separators: 

separators

These 3 are similar to the default, wide and dots separators and should probably overwrite them.

A second option would be to create these 3 as custom block styles.

Currently, the default style is identical to the wide style, since it is only styled as a hr, not with the class `wp-block-separator`.

### LINKED ISSUE #255: no style for nested emphasis
Does the design consider nested `em`, `i`, `cite` tags?
Please fill in this style:
`em em, em i, i em, i i, cite em, cite i { }`

andersnoren:
Includes a bunch of updates to the block-styles on the front-end (no editor styles).

## Block styles:
- Base block margins
- Pullquote
- Blockquote
- Widget blocks (Latest Posts, Archive, Recent Comments)
- Group block
- Separator block (fixes #254)

The block styles have been tested with the updates to the `.alignleft` and `.alignright` classes included in #236 and #237.

## Also:
- Added styling for nested emphasis (fixes #255)
- Removed removal of outline from base input styles (fixes #67, inputs with their own focus styles might need updating)
- Added styles for `big` and `small`, adjusted size of `code` (fixes #88)

### DIFF
--- a/style.css
+++ b/style.css
@@ -381,6 +381,10 @@ dfn {
 	font-style: italic;
 }
 
+em em, em i, i em, i i, cite em, cite i {
+	font-weight: bolder;
+}
+
 b,
 strong {
 	font-weight: 700;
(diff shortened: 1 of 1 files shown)


carolinan:
How thorough do you want us to test the block styles before we merge it?

andersnoren:
@carolinan Not sure. I've tested them with all of the theme unit test data and added a couple of custom test pages for the more style heavy/missing blocks (pull quotes, groups and so on), but it's likely that I've missed something. I think it might be better to get them in as early as possible, so we get more eyes on them more quickly.

style.css
```diff
@@ -385,6 +385,10 @@ em em, em i, i em, i i, cite em, cite i {
 	font-weight: bolder;
 }
 
+big { font-style: 1.2em; }
+small { font-style: .75em; }
```
joyously: There's no such property as a numeric font-style. I think you mean font-size.

PR #271, opened 2019-09-14
UPDATE #270: Use } instead of endif for consistency reasons

### LINKED ISSUE #270: Use } instead of endif for consistency reasons
The files `header.php` and `index.php` currently contain several `endif` statements instead of closing brackets. While this is fine according to https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style due to consistency reasons, we should go with one style. As the rest of the theme is using `}` I'll change the two files accordingly. PR is on the way. 😁

nielslange:
Fixes #270


nielslange:
Closed in favour of #272

Issue #267, opened 2019-09-14
Consider using arrays for $selector in twentytwenty_generate_css


celloexpressions:
`twenty_twenty_generate_css()` currently accepts text strings for the `$selector` parameter. In many of the current uses of this function within the theme, the selector list includes several comma-separated selectors.

Formatting the selector as an array, rather than a string, would make it easier to filter the selectors passed to this function (and reduce the risk of missing or extra commas in this final CSS).  This approach would also improve code readability.

It could also work to accept either a string or an array. I would favor arrays when three or more selectors are used, whereas a string is reasonable for only one or two selectors. `twentytwenty_get_customizer_css()` already has several calls to this function with three or more selectors.

acosmin:
@celloexpressions I tried something 😄  I might have overdone it

PR #273, opened 2019-09-14
Filterable selector for `twentytwenty_generate_css()`

acosmin:
Fix for #267

### DIFF
--- a/inc/custom-css.php
+++ b/inc/custom-css.php
@@ -32,10 +32,17 @@ if ( ! function_exists( 'twentytwenty_generate_css' ) ) {
 			$selector = array_map( 'trim', explode( ',', $selector ) );
 		}
 
-		$selector = apply_filters( 'twentytwenty_generate_css_selector', $selector );
 		$selector = implode( ', ', array_unique( $selector ) );
 
-		$return = sprintf( '%s { %s: %s; }', $selector, $style, $prefix . $value . $suffix );
+		$args = compact( 'selector', 'style', 'value', 'prefix', 'suffix' );
+		$args = apply_filters( 'twentytwenty_generate_css_args', $args );
+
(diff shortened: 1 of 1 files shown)


inc/custom-css.php
```diff
@@ -28,7 +28,23 @@ function twentytwenty_generate_css( $selector, $style, $value, $prefix = '', $su
 			return;
 		}
 
-		$return = sprintf( '%s { %s: %s; }', $selector, $style, $prefix . $value . $suffix );
+		if ( ! is_array( $selector ) ) {
+			$selector = array_map( 'trim', explode( ',', $selector ) );
+		}
+
+		$selector = apply_filters( 'twentytwenty_generate_css_selector', $selector );
```
celloexpressions: Can we come up with a good way to add context to this filter? For example, we could move the filters in `twentytwenty_get_customizer_css`, such as `apply_filters( 'twentytwenty_buttons_targets_front_end', ...`, to make the generated CSS universally filterable from the `generate_css_selector` filter. We need a way to pass the `buttons_targets_front_end` context to this new filter, then we can remove the redundant filter in the theme.

celloexpressions:
This looks like a good start. I added one specific comment. We should also include inline filter docs to match the core inline docs patterns.

inc/custom-css.php
acosmin: We could probably add a new function argument `$context` and use it in the filter name or a second param...

`apply_filters( 'twentytwenty_generate_css_selector', $selector, $context )`
or
` apply_filters( 'twentytwenty_generate_css_selector_' . $context, $selector )`

joyously: Or just remove the function completely and use CSS Custom Properties for those things that need generating.

Issue #279, opened 2019-09-15
Close icon for search extends outside of mobile view


okpierre:
Here's the screen capture showing the X icon on the right side getting cut off

![twentytwenty-mobile-search](https://user-images.githubusercontent.com/1679025/64915226-a51c2d00-d730-11e9-9858-4c893ab50fc8.png)

andersnoren:
Fixed by #280

carolinan:
@okpierre what is your wordpress.org username?

PR #280, opened 2019-09-15
Search untoggle fix - mobile

acosmin:
Fix for #279


acosmin:
seems to work now :) good catch...

PR #281, opened 2019-09-15
Make the theme use minified files, introduce Grunt

jdevalk:
I am using the theme in production on [joost.blog](https://joost.blog) and found that the CSS and JS files could use some optimization. In order to do that effectively, I think we should introduce `grunt` as a task loader.

## Grunt tasks
I've made a `Gruntfile.js` with the following tasks:

- `build:js` - this currently only minifies the JS through grunt-uglify
- `build:css` - this copies the `style.css` file and `assets/css/font-awesome.css` file to `assets/dist/css` and then runs them through `postcss`. This in turn runs `cssprefixer` on them to prefix for specific browsers where needed, and then runs `cssnano` on it to minify the CSS.
- `build` - which combines the above two tasks.
- `watch` - this can be run using `grunt watch` on the command line as you're editing CSS and/or JS and it will run both `build:js` and `build:css` on the fly as needed.

## Loading minified files
The theme now loads minified files all the time. In order to ease development, both the JS and the CSS files have source maps so you can easily find the proper lines in the unmodified source files.

I've taken the liberty of updating `CONTRIBUTING.md` in this pull to reflect the above.


celloexpressions:
This significantly increases the barrier to entry for contributors and for users that want to start digging into how their site works. While there are benefits to optimization, the costs to doing this are far greater. The complexity of build tooling should never be used in a default theme.

I would be open to the use of minification only if it if done in an automated way as part of the final release process, understanding that this still has negative impacts on site owners.

jdevalk:
I disagree with that @celloexpressions. Default themes should show what WordPress is capable of, and of course the build tools we use within them should be well documented. We should not stay away from them with reasons like "it's too hard", that's sending the wrong message about what WordPress is capable of achieving.

ianbelanger79:
I am going to close this for now, but I think we might want to revisit this before Beta 1, if we have time. @jdevalk makes some good points on why we should use build tools and I tend to agree.

afercia:
> The complexity of build tooling should never be used in a default theme.

Twenty Nineteen has several build scripts. Amongst other things, it introduced the usage of .scss files in the an official Bundled Theme. Building is necessary. Twenty Twenty uses .scss files as well. 

I don't see a good reason why Twenty Twenty should go in the opposite direction taken by Twenty Nineteen.

joyously:
> Twenty Twenty uses .scss files as well.

No it doesn't.
It's not about going in any particular direction. It's about how much time there is to work on the theme, and the experience level of the team.

afercia:
> No it doesn't.

Ouch, sorry I was convinced it did, for some reason. Maybe I dreamed about it.

PR #282, opened 2019-09-15
Update the header and menu modal with the conditionals for the new header structure

### LINKED ISSUE #27: A11y: display close button in the menu panel
I think there should be a close button in the menu panel, it would be useful for users that use their keyboard to navigate through the content of the website.

I know the menu button toggles to a close button when it's clicked, but the toggled close button should probably be displayed on the menu panel so the user could see the focus is on a close button :-)

### LINKED ISSUE #93: Icon-only controls are not accessible
In the current version of the theme c277a895704cc00ec402261a7f30fdd409d03ffa the Search and Menu controls in the header use both visible text and icons:

Screenshot 2019-09-07 at 19 38 34

These come from [the Chaplin theme](https://wordpress.org/themes/chaplin/) and are perfectly usable by all users.

Instead, in the design shown in [the Twenty Twenty post on Make](https://make.wordpress.org/core/2019/09/06/introducing-twenty-twenty/) the header design is different and uses icon-only controls. This new design is going to be implemented in https://github.com/WordPress/twentytwenty/pull/71

header order

However, icon-only controls are not accessible. [Icons don't have a universal meaning](https://www.nngroup.com/articles/icon-usability/). Not all users can operate on UI controls that don't expose their accessible name visually.

The same issue was discussed at length in the Gutenberg project. "icon-only" controls are problematic for many reasons. They're an accessibility anti-pattern as pointed out in the first [Gutenberg accessibility report](https://make.wordpress.org/accessibility/2018/10/29/report-on-the-accessibility-status-of-gutenberg/) published by the accessibility team on October 2018.

For example, speech recognition software users need to know the accessible name (the label of control's text) to be able to voice a command the software can understand, as well explained in https://github.com/WordPress/gutenberg/issues/470#issuecomment-435611115

A few previous related comments on the Gutenberg repository:
- https://github.com/WordPress/gutenberg/issues/5358#issuecomment-371070915
- https://github.com/WordPress/gutenberg/issues/7221#issuecomment-398748271

Similarly, the recent WPCampus Gutenberg accessibility audit by Tenon pointed out the same issue in various part of the WordPress interface. See for example:

Visible labels are not available on editing buttons
https://github.com/WordPress/gutenberg/issues/15352
then moved to Trac https://core.trac.wordpress.org/ticket/47116

Icons alone are visible for controls unless they are hovered or focussed
https://github.com/WordPress/gutenberg/issues/15311

In both cases, the provided remediation guideline is to add visible text labels to the buttons or a setting so that users can specify whether to show icons-only, labels-only, or icons and labels together. For what concerns Gutenberg, the second option is currently explored in a 
 pending pull request, see https://github.com/WordPress/gutenberg/issues/10524 and https://github.com/WordPress/gutenberg/pull/15830. I do realise user settings on a theme would be uncommon. The alternative is to add visible text.

andersnoren:
This PR builds on the menu location updates @carolinan made in #247, and updates it with the conditionals and fallbacks specified in #246. It also updates the style to match the design.

I couldn't find a way to replicate the `wp_nav_menu` sub menu button toggles and chevron icons in `wp_list_pages` without including a custom walker that modifies `start_el`. If we can find a way to do this without using a custom walker, that would be much, much better.

In addition to the things in that issue, it includes:
- Text labels for the header toggles (fixes #93)
- Close button in the modal menu (fixes #27)


pattonwebz:
I am seeing a little strangeness on the admin bar when search/menu appears.

Peek 2019-09-15 14-54

Also a couple clicks on menu ends up putting it into an _open but not visible_ state for some reason.

andersnoren:
@andersnoren I've updated the modal styles to include a top offset equal to the height of the `.admin-bar`, but it won't work until the JavaScript is updated with a different scroll lock method. Since the header toggles are always at the top of the screen, we can simply scroll to the top and set the `body` to `overflow: hidden` when a modal is shown, at which point the modals will be positioned correctly.

As for the toggles, yes, the JavaScript handling that should most definitely be rewritten as well :)

pattonwebz:
@andersnoren did you mean to mention yourself or were you trying to mention me? lol I am going to look a bit at javascript stuff now so will try see if we can resolve some of that before end of day :)

andersnoren:
@pattonwebz A bit quick there 😛 Awesome!

joyously:
> I couldn't find a way to replicate the wp_nav_menu sub menu button toggles and chevron icons in wp_list_pages without including a custom walker that modifies start_el.

I've had this problem before and I'm sure a lot of theme authors have this problem. Could we make a patch for core with a filter or something in the right place so we don't all have to write custom walkers?

pattonwebz:
So I tested the javascript PR at #163 and it does seem to be working. I was unable to properly test merging it with this PR but I think there are some conflicts we need to spend time tomorrow working though so we can get both this and the javascript swap PR into master

PR #283, opened 2019-09-15
Add a customizer option for showing excerpts in archives.

### LINKED ISSUE #2: Add option to use excerpts on posts page
The Chaplin theme has an option to use excerpts on the default posts page. This option is currently absent from Twenty Twenty. I think it would be a good idea to bring back this option from Chaplin. In fact, I'd prefer if excerpts were used by default. The inability to change this has been a major missing feature in previous default themes, in my opinion.

carolinan:
Adds a customizer option for showing excerpts in archives.
Renames the customizer section to Theme Options
Fixes #2

### DIFF
--- a/404.php
+++ b/404.php
@@ -14,9 +14,9 @@ get_header();
 
 	<div class="section-inner thin">
 
-		<h1 class="archive-title"><?php esc_html_e( 'Page Not Found', 'twentytwenty' ); ?></h1>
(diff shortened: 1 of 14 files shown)


celloexpressions:
This solves two of my comments: that customize sections should typically limit themselves to about 1 theme-specific section and that sections should generally contain multiple controls.

It looks like this may currently have several other PRs applied.

carolinan:
> It looks like this may currently have several other PRs applied.

Yes something went wrong here. Those other PR's are already applied...

PR #98, opened 2019-09-07
ADD #43: Add screenshot

nielslange:
Fixes #43 

<table>
<tr>
<td>Before:
<br><br>
43 - before

</td>
<td>After:
<br><br>
43 - after

</td>
</tr>
</table>

@andersnoren Feel free to replace screenshot.png with a more suitable one if you like to. 😉

PR #299, opened 2019-09-16
Added screenshot, updated readme.txt with license

andersnoren:
Added a screenshot and updated readme.txt with licensing information. Opted for a jpg version to reduce file size (the png version of the same screenshot was 600 KB after compression).

screenshot

Closes #98.


acosmin:
Seeing how this shouldn't be yet another blog theme :) wouldn't it be more effective to use a screenshot like you have for Chaplin

andersnoren:
@acosmin Yeah, the screenshot might be worth revisiting later, to replace this one with a screenshot that show off the blocks better. I'm merging this in for now though, so we have a screenshot in there. :)

Issue #301, opened 2019-09-16
Fix translations in credits.


juanfra:
Currently, we're not translating the URL to WordPress but the doc for translators says that.

We should be making the URL to WordPress translatable, so each locale can have their own link. And at the same time being consistent with the doc for translators. Not sure if the usage of `_x()` is correct there.

pattonwebz:
Resolved by: https://github.com/WordPress/twentytwenty/commit/eec0bc0d4d89a5a7ee6aab1509164f516504890a

pattonwebz:
Thanks for pointing this out and for the PR. I merged that in just there so this is now in line with past core themes.

juanfra:
Thank you, William! :)

PR #302, opened 2019-09-16
Fix translations in credits.

juanfra:
Currently, we're not translating the URL to WordPress but the doc for translators says that.

We should be making the URL to WordPress translatable, so each locale can have their own link. And at the same time being consistent with the doc for translators. Not sure if the usage of `_x()` is correct there.

**Fixes:** #301 
**WP Username:** juanfra


footer.php
```diff
@@ -131,10 +131,12 @@
...
 
 						<p class="powered-by-wordpress">
-							<?php
-							/* Translators: %s = Link to WordPress.org */
-							printf( esc_html_x( 'Powered by %s', 'Translators: %s = Link to WordPress.org', 'twentytwenty' ), '<a href="https://wordpress.org">' . esc_html__( 'WordPress', 'twentytwenty' ) . '</a>' );
-							?>
+							<a href="<?php echo esc_url( __( 'https://wordpress.org/', 'twentytwenty' ) ); ?>">
+								<?php
+								/* Translators: %s: WordPress. */
+								printf( __( 'Powered by %s.', 'twentytwenty' ), __( 'WordPress', 'twentytwenty' );
```
pattonwebz: Missing closing bracket here

juanfra: Oh sorry for that!

juanfra:
Perfect, thanks :)

We can either wait for a decision on https://github.com/WordPress/twentytwenty/pull/177 or I can update the code to use the escaped version.

pattonwebz:
In line with the other PR about similar translation escapes I expect that whitelisting the missing escape with `// phpcs:ignore` comment is the best option here.

juanfra:
Thank you @pattonwebz ! Perfect, I updated it.

PR #308, opened 2019-09-16
UPDATE #307: Increase line length

### LINKED ISSUE #307: Line length seems too narrow
The line length which is defined by the combination of the text width (58rem) and the font size (2.1rem on desktop) seems too narrow. The number of characters per line is in the low range of the ideal range. Text inside 2 columns is also too narrow (by default the columns block width is set to 58rem). Imho, I think the readability can be improved by increasing the number of characters per line.

Increasing the text width from 58rem to 68rem could improve readability without changing the overall feeling of the theme. I’ve built page layouts using blocks (group, columns, media & text) and I’ve found that the line length of Chaplin is more versatile. 68rem would give a line length quite similar to Chaplin.
  
@andersnoren What do you think?

nielslange:
Fixes #307 

<table>
<tr>
<td valign="top">Before:
<br><br>

#307-1 - before

</td>
<td valign="top">After:
<br><br>

#307-1 - after

</td>
</tr>
</table>

### DIFF
--- a/style.css
+++ b/style.css
@@ -861,7 +861,7 @@ section {
 	width: 100%;
 }
 
-.section-inner.thin { max-width: 58rem; }
+.section-inner.thin { max-width: 68rem; }
 .section-inner.medium { max-width: 100rem; }
 
 .section-inner.no-margin { margin: 0; }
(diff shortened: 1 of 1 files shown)


andersnoren:
Sorry @nielslange, closed for the same reason as given in #308.

Issue #300, opened 2019-09-16
Use core functionality for single navigation


yannickiki:
Core has a function `the_post_navigation()` to display the navigation to next/previous set of posts. It implements screen reader text for navigation elements.

nielslange:
@yannickiki What's wrong with the current implementation? As far as I know, the `aria-label` in https://github.com/WordPress/twentytwenty/blob/ef21a6df7125a84dcfdcf61bdd4119d6a80331ef/template-parts/navigation.php#L25 is interpreted by screen readers to tell users what element this is. Having said that, I think it's a good idea to adjust that label from `Post` to `Post navigation`. I'll create a PR for that the next minutes.

cc: @afercia

yannickiki:
@nielslange I thought that themes should use core functionalities if they exist.

nielslange:
@yannickiki I created #309 to adjust the `aria-label` and send a PR with https://github.com/WordPress/twentytwenty/pull/310.

nielslange:
> @nielslange I thought that themes should use core functionalities if they exist.

Good point, @yannickiki. That's probably an issue someone from the theme review team can clarify.

cc: @carolinan, @pattonwebz

afercia:
>  I think it's a good idea to adjust that label from Post to Post navigation

Please don't 🙂 Screen readers already announce the `<nav>` element type as "navigation". The aria-label should not contain the word "navigation" again, to avoid it gets repeated unnecessarily:

<img width="1310" alt="Screenshot 2019-09-16 at 18 27 27" src="https://user-images.githubusercontent.com/1682452/64975634-c7888480-d8af-11e9-8309-447d36b6662c.png">

nielslange:
@afercia Do I understand you correctly that screen readers first read the `aria-label` and then the element? Thus, a screen reader would read `Post, navigation` automatically, right?

nielslange:
Closed based on https://github.com/WordPress/twentytwenty/issues/309#issuecomment-531854465

afercia:
@nielslange yes: the order may vary across screen readers and browsers combinations and also depends on the type of element (links, buttons, form fields) but all screen readers announce both the label (aria-label, etc.) and the element type.

nielslange:
Thanks for explaining this to me, @aferica. I’ve already closed the PR that i opened for that.

PR #314, opened 2019-09-16
Using core's post navigation and pagination functions

anace:
This PR solves issue #300 (use core functionality for single navigation). Core's post navigation is accessibility ready and simpler to use. Besides, one of the Team Review Theme requirements is to use core functions when they exist.

In my PR, I kept the old code, both in `navigation.php` and in the stylesheet, so the original navigation and the new can be compared on the browser. There are indications on the two files around the code that has to be deleted in case the request is accepted.

WordPress username: acalfieri


pattonwebz:
In the case of the single post navigation styles for both approaches look the same in my browser. Markup is a bit different.

Upper = original, Lower = addition in this PR.
Screenshot from 2019-09-18 02-18-13

Upper:
```
<nav class="pagination-single section-inner" aria-label="Post">

		
			<a class="previous-post" href="http://local.themereview/2009/05/edge-case-nested-and-mixed-lists">
				<span class="arrow">←</span>
				<span class="title"><span class="title-inner">Edge Case: Nested And Mixed Lists</span></span>
			</a>

			
			<a class="next-post" href="http://local.themereview/2009/07/edge-case-many-categories">
				<span class="arrow">→</span>
					<span class="title"><span class="title-inner">Edge Case: Many Categories</span></span>
			</a>
			
	</nav>
```
Lower:

```
    <nav class="navigation post-navigation" role="navigation">
		<h2 class="screen-reader-text">Post navigation</h2>
		<div class="nav-links"><div class="nav-previous"><a href="http://local.themereview/2009/05/edge-case-nested-and-mixed-lists" rel="prev"><span class="arrow" aria-hidden="true">←</span><span class="screen-reader-text">Previous post:</span><span class="post-title">Edge Case: Nested And Mixed Lists</span></a></div><div class="nav-next"><a href="http://local.themereview/2009/07/edge-case-many-categories" rel="next"><span class="arrow" aria-hidden="true">→</span><span class="screen-reader-text">Next post:</span><span class="post-title">Edge Case: Many Categories</span></a></div></div>
	</nav>
```

Posts with no titles are handled better with the `the_post_navigation()` 

Screenshot from 2019-09-18 02-15-16

carolinan:
How does it handle the styling for when there is only one arrow? On the first and last page.

carolinan:
The numbers are meant to be centered:

numbers

See https://github.com/WordPress/twentytwenty/issues/99

carolinan:
Below 1000 pixels the spacing is not correct:

numbers2

carolinan:
I am still super confused. Can these be split into two separate pull requests, one for each navigation,
and then remove the CSS that would no longer be used?

carolinan:
@afercia Is using the screen reader text from the default markup acceptable here? Or does it need to be an aria label? https://github.com/WordPress/twentytwenty/issues/73

afercia:
@carolinan good question and thanks for the ping! There's a core Trac ticket aimed to add the missing `aria-label` to the various post(s) / comments navigation. See:

The post(s) navigation landmark needs to be labelled
https://core.trac.wordpress.org/ticket/47123

Hopefully will be committed soon in core. Some testing would be nice 🙂

carolinan:
Since the label hopefully will be included in 5.3, that's one less reason for using the custom markup.

anace:
If you accept the change, I can continue working with the styles, so it looks like in the design.

carolinan:
This PR cant be merged, it needs to be updated and any duplicate code and CSS must be removed.

anace:
I'll close this PR and start from the beginning.

PR #65, opened 2019-09-07
Add RTL support

senadir:
this PR add support for RTL by adding rtlcss & running it.

It also adds the npm CLI command `npm run build:rtl` to generate a new rtl file.

It also adds the RTL file in function so it can be loaded when needed.

## How to test
* after pulling this PR, run `npm install` to install dependencies, it should be noted that you must have both [node & npm](https://nodejs.org/en/download/) (npm is bundled with node) installed in your machine, [this tutorial might help you](https://javascriptforwp.com/how-to-install-npm-and-node-on-a-mac/)
* run `npm run build:rtl` to generate the new RTL file
* either switch the site language to an RTL one (arabic, hebrew, urdu, persian...) or install this handy plugin [RTL tester](https://wordpress.org/plugins/rtl-tester/)
* verify that everything is well aligned.

### note
there is a big about an unflipped arrow in the post pagination section.

PR #316, opened 2019-09-17
RTL stylesheet rebuild

pattonwebz:
This PR rebuilds the RTL stylesheet after #65 was merged and sets the theme to output it.


pattonwebz:
@senadir if you could review this too that would be great. It's just running the new build task and then adding the code to output frontend.

Issue #323, opened 2019-09-17
Add separators between posts 


carolinan:
In the design, archives has a separator between each article.
The separator uses the same design as the **_wide line separator block._**

![separators2](https://user-images.githubusercontent.com/7422055/65011908-96429f80-d915-11e9-8330-00e8ea827398.png)

carolinan:
Merged, closing

PR #329, opened 2019-09-17
Add separators to archive pages, update pagination

### LINKED ISSUE #99: Add styles to the newer / older post pagination

andersnoren:
Includes separators between posts on archive pages using `<hr />` elements, to simplify shared styles between the entry content separators. I've set the separators to `aria-hidden="true"` – please let me know if that should be changed to something else.

I've also added a matching separator to the pagination, and tested a solution for the problem with the items not being centered when there's no "Older Posts" or "Newer Posts" links being output. The solution either prepends of appends a placeholder element matching the missing one, and sets it to `aria-hidden="true"` and `visiblity: hidden` (not sure if both are needed). It's not a very neat solution, but it gives us consistent spacing between the items in the pagination regardless of which page we're on.

Fixes #99.


carolinan:
When I test this in IE11, there is a large horizontal scroll when the width is over 1000px.

andersnoren:
There's an issue in the site header in IE11. I'm going to try and find a fix for it tomorrow morning.

carolinan:
Was the nav for https://github.com/WordPress/twentytwenty/issues/73 removed in an earlier PR? I don't see it as part of this.

ianbelanger79:
Will fix #323

andersnoren:
> There's an issue in the site header in IE11. I'm going to try and find a fix for it tomorrow morning.

Added a fix for this in #359.

carolinan:
Merged with a reservation for #73 being fixes as soon as possible.

Issue #38, opened 2019-09-07
Add/improve starter content


carolinan:
Twenty Nineteen did not include starter content, and we could debate its usefulness since it only works on fresh sites.

We can use the starter content as a way to help the user go from the pretty but blank theme, to a theme that does showcase blocks for users on fresh sites.

This is a nice to have feature if we have the time but I also think it could be really good.

LittleBigThing:
See [here an attempt](https://github.com/LittleBigThing/twentynineteen/blob/09150cf2d750d3122be2104eb086914166d03049/inc/template-functions.php#L428) to add block-based starter content to Twenty Nineteen. It might be a starting point, although it should be at least updated to the most recent block markup.

celloexpressions:
Starter content is absolutely necessary here because this is the default theme. The vast majority of sites will only ever see any starter content if it's included with Twenty Twenty.

The eventual goal of the starter content feature is to also work on sites with existing content. More themes (especially bundled themes) using the feature would help create momentum toward improving the feature in core.

joyously:
The starter content could be copied from Twenty Seventeen, to start.

carolinan:
The plan is to get any stater content in for beta 1, then iterate on it.
The second part of the plan is to include blocks to make a homepage that matches the design mockup.

garrett-eclipse:
From a discussion on Slack here - https://wordpress.slack.com/archives/C02RQBWTW/p1570726252294400

It was flagged the 2020 starter content for the homepage isn't really generic or reusable.

joyously:
I say remove the page content about the Museum and use the core-defined pages instead.

garrett-eclipse:
I would second that @joyously especially if we can land core#48130 into 5.3 and have it all be block enabled. Only concern is it showcases less of the block capabilities as it's just a paragraph block instead of the multiple blocks found in the museum content

joyously:
I think it's more important that the starter_content be useful and not a hassle for the user (to delete or have to change much). The intent is not to showcase the editor or blocks, but to get the user's site started easy and fast. They shouldn't have to change much, because starter content is all or nothing (can't accept only some parts of it).

PR #334, opened 2019-09-17
Add some starter content to the theme

pattonwebz:
Adds a first set of starter content. Very basic, does not provide any images, blocks or other post content.

This starter content is loosely based on the set used in Twenty Seventeen. For some inspiration for blocks see this comment: https://github.com/WordPress/twentytwenty/issues/38#issuecomment-529178328


joyously:
Hey, I was answering a forum question about using the Latest Posts block in the static Home Page.
I suppose that's one way to do starter content that's fairly generic. (but it would only show 1 post)
Or would that be too confusing?

carolinan:
The menu names need to be updated and the social menu only seem to work in the modal, not in the footer, but we can update that when the rest of the starter content is added.

Issue #335, opened 2019-09-17
Site title won't display in customizer


pattonwebz:
The logic used to display the site title is based on if `get_the_logo()` returns a value or not.

However `get_the_logo()` always returns a value in the customizer even when no logo is set. This is because a placeholder is needed for selective refresh when logo is added/removed.

This causes the site title to never display in the customizer because the test on `$logo` is always true.

ianbelanger79:
Fixed in #355

PR #336, opened 2019-09-17
Detect when `$logo` is only a placeholder and display title in that case

pattonwebz:
This PR tries to detect when `get_the_logo()` returns a placeholder rather than a real logo. Placeholders are returned in the customizer when no logo is set.

The placeholder has `display:none;` property so I am checking for that with `strpos`. Other test methods may be more suitable but this works for now.

Fixes: #335


joyously:
I think a lot of themes output the title to fit in the space of the logo. If the logo is there, it is on top of the title. 
But that wouldn't work well for logos with transparency, or lots of long words.
Does the logo have to supplant the title? A user can't show both?

pattonwebz:
@joyously there may still be some old school thinking behind why that is common along the lines of _if the logo image doesn't load the title still shows'.

In this theme though when logo is present we use the site title as a screen reader text instead.
https://github.com/WordPress/twentytwenty/blob/6e5dd9d703cf52db7bcc0d550673cc09b3e5fd60/header.php#L73

ianbelanger79:
Thanks for tackling this @pattonwebz. It does work when you load the customizer and do not have a Logo set, but if you set a logo and then remove that logo, it reverts back to the previous issue, where there is no site title. I would not focus on this right now though. We can take care of it after beta 1.

acosmin:
@pattonwebz This can be simply solved by adding a partial, like

**functions.php** add:
```
function twentytwenty_site_logo() {
	$logo       = get_custom_logo();
	$site_title = get_bloginfo( 'name' );

	if ( has_custom_logo() ) {
		$home_link_contents = $logo . '<span class="screen-reader-text">' . esc_html( $site_title ) . '</span>';
		$site_title_class   = 'site-logo';
	} else {
		$site_title_class   = 'site-title';
		$home_link_contents = '<a href="' . esc_url( home_url( '/' ) ) . '">' . esc_html( $site_title ) . '</a>';
	}

	if ( is_front_page() || is_home() ) {
		echo '<h1 class="' . esc_attr( $site_title_class ) . '">' . $home_link_contents . '</h1>'; //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Escaped during generation.
	} else {
		echo '<div class="' . esc_attr( $site_title_class ) . ' faux-heading">' . $home_link_contents . '</div>'; //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Escaped prior to this point.
	} 
}
```

**classes/class-twentytwenty-customizer.php:**
```
$wp_customize->selective_refresh->add_partial(
    'custom_logo',
    array(
        'selector'        => '.header-titles [class*=site-]:not(.site-description)',
        'render_callback' => 'twentytwenty_site_logo',
    )
);
```

**header.php - change L60-79 to**:
```
<?php
    twentytwenty_site_logo();
   $site_description = get_bloginfo( 'description' );
```

ianbelanger79:
Fixed in #355

Issue #285, opened 2019-09-15
Instant live preview of color customization


celloexpressions:
Instant live preview of color customizations in the customizer can be a great experience and help users quickly select just the right colors for their site. There are several potential implementation approaches. Ideally, the solution will:
1. Instantly preview most of the colors in JS.
2. Not duplicate any CSS logic in PHP and JS.
3. Not duplicate color calculation logic in PHP and JS.
4. Ensure that previewed color CSS exactly matches what's shown on the frontend.

### Proposal
I am proposing a general strategy rather than an actual PR so that we can make decisions about the approach before building an implementation following decisions on #284.

Starting with item 4, the current approach performs a full page refresh in the customize preview to preview colors and match the frontend. We can refine this behavior by leveraging selective refresh, only reloading the relevant `style` element from PHP. Conceptually, this is done with:

```
$wp_customize->selective_refresh->add_partial( 'twentytwenty_colors', array(
	'selector'        => '#twentytwenty-colors', // <style> element
	'settings'        => array( 'background_color', 'accent_color' ),
	'render_callback' => 'twentytwenty_color_css_frontend',
) );
```
To implement item 1 while observing 2 and 3, we need to creatively consider ways to update some of the colors without duplicating logic, knowing that the full CSS will be rebuilt with color calculations in PHP periodically. I've found that doing a find/replace for the customize color within the generated CSS works well for this purpose. 

This starts with adding the color setting values as `data-` attributed to the `<style>` element if `is_customize_preview()`. Then, our settings are set to `transport => postMessage`. Finally, in the enqueued customize preview JS file, we can use something like this for each color setting:

```
wp.customize( 'accent_color', function( value ) {
	value.bind( function( to ) {
		// Update custom color CSS
		var style = $( '#twentytwenty-color' ),
		      color = style.data( 'accent-color' ),
		      css = style.html();
		css = css.split( color ).join( to ); // equivalent to css.replaceAll.
		style.html( css )
		       .data( 'color', to );
	} );
} );
```

Does anyone have alternative approaches to consider for discussion?

joyously:
CSS Custom Properties

aristath:
While I agree that duplicating logic in PHP & JS is sub-optimal, using partial-refreshes won't be less code, and the result won't be instant in the customizer, there will be delays.

Alternative approaches:

[...]

PR #341, opened 2019-09-18
Selective Refresh for colors

acosmin:
Based on #285

Adds selective refresh for colors. 

The only color that doesn't work is `cover_template_overlay_background_color` because for some reason it has the background color added inline. Should I change that one too?

Also, the js is ES5 but 😄 it wasn't really needed, jQuery is a major dependency in the Customizer.


carolinan:
Looking at the discussion in the issue, I don't believe a finally desicion has been made about this yet. It is probably best to wait until the color settings are updated.

acosmin:
Yep, in case nothing happens in that direction there's always this. Even though I think we should let the user select what colors he wants and in which combos he desires...

acosmin:
closing based on #351

PR #350, opened 2019-09-18
Using core's navigation function for single post navigation

anace:
This PR solves issue #300, as in PR #314. This time I have only changed single post navigation and deleted the existing code.


anace:
I close this PR and will make a new one from the updated master files.

Issue #284, opened 2019-09-15
Audit Color Combinations and Automatically Correct Color Contrast at User-Customizable Colors


celloexpressions:
There have been several issues and PRs related to color options. Now that the initial implementation is mostly complete, let's evaluate whether Twenty Twenty provides an appropriate amount of customization without over-complicating the user experience, in accordance with the WordPress Philosophy.

We also need to review all combinations of text and background colors for:
1. Color contrast ratios with the default colors.
2. Maintaining color contrast where one of the colors is user-customizable.
3. Maintaining a minimum contrast threshold when both of the colors are user-customizable (by adjusting the selected color(s) as required).

Ideally, the result will be that users can change all of the colors in the theme and that a minimum color contrast threshold will be maintained regardless of the user's combination of selected colors. The current default color palate is:

#fff - header/footer background
#f5efe0 - body background
#dcd7ca - borders
#cd2653 - accent
#6d6d6d - secondary text
#000 - text

![image](https://user-images.githubusercontent.com/7875961/64923022-ac702480-d78a-11e9-84b6-ace7e8988945.png)

### Initial Proposal

Based on my audit of color usage, I would propose the following adjustments:
1. Merge the cover template overlay background color option with the accent color option. This option defaults to the accent color already, and is likely workable with the selected accent color for a large percentage of users. The separate option doesn't pass the initial 80/20 test for "Decisions, not Options".
2. Eliminate or simplify the cover template overlay text color option in favor of automatically-determined white or black. The image background makes it too difficult to determine actual contrast, but we can determine whether white or black will work best based on the combination of accent color and opacity. If we decide that this option needs to stay, we may want to simplify it to a preselected  list of colors instead of an open-ended color picker. Without this adjustment, users will likely lean too heavily on an obscure text color rather than selecting an appropriate level of opacity for contrast. Without a text color option, users would have to adjust opacity to ensure visible readibility.
3. Add an option to customize the header/footer background color. This is a big missing piece currently, and makes the theme substantially more flexible.
4. Don't add any text color options as proposed in #3.

With these adjustments, users can customize these colors:
1. Hader/footer background color.
2. Primary background color.
3. Accent color.

And the theme would "calculate" selections for these colors when the defaults don't provide adequate contrast (via an extension of #249):
1. Header/footer text color.
2. Body text color.
3. "Secondary" text color (default #6d6d6d) in the header/footer.
4. "Secondary" text color (default #6d6d6d) in the body.
[...]

PR #351, opened 2019-09-18
See #284

aristath:
This addresses #284. Not to be merged yet, this is a proof of concept. What it does: (see screencast below)

demo2

The implementation here replaces the PHP class that was already added in #249 and that is no longer needed, if this get merged (after we tweak it) then the other class can be removed.

I'll leave comments in the code so we can review things and decide what to do.

For now I only did this for the content area but the code can handle as many areas as we want, so it can different for the header, footer, overlays etc.


assets/js/color.js
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @return {Object} - this
+ */
+_twentyTwentyColor.prototype.setAccentColorsArray = function() {
+	var minSaturation    = 40,
+		maxSaturation    = 100,
+		minLightness     = this.isDark ? 40 : 0,
+		maxLighness      = this.isDark ? 100 : 60,
+		stepSaturation   = 4,
+		stepLightness    = 4,
```
aristath: We can tweak these variables here to change the accuracy of the script and how it calculates things.

assets/js/color.js
```diff
@@ -0,0 +1,132 @@
...
+		rules = [
+			[ 7, 3 ],
+			[ 6, 3 ],
+			[ 5, 3 ],
+			[ 4.5, 3 ],
+			[ 4.5, 2 ],
+			[ 4, 3 ],
+			[ 4, 2 ],
+			[ 3, 3 ],
+			[ 3, 2 ]
```
aristath: Perhaps we should tweak this array?

assets/js/customizer-preview.js
```diff
@@ -0,0 +1,49 @@
...
+
+	var backgroundColors = { // TODO: Define in PHP then add as vars to the script.
+		content: {
+			setting: 'background_color',
+			elements: {
+				text: 'body',
+				accent: 'a, .wp-block-button.is-style-outline, .has-drop-cap:not(:focus):first-letter, a.previous-post, a.next-post'
+			}
+		}
+	};
```
aristath: This should be moved to a `wp_localize_script()` call, both for this file and for the file responsible for the controls. We can then add a filter so users can add their own.

assets/js/customizer.js
```diff
@@ -0,0 +1,64 @@
...
+		// TODO: Move these to PHP and add as vars to the script.
+		var backgroundColors = { // TODO: Define in PHP then add as vars to the script.
+			content: {
+				setting: 'background_color',
+				elements: {
+					text: 'body',
+					accent: 'a, .wp-block-button.is-style-outline, .has-drop-cap:not(:focus):first-letter, a.previous-post, a.next-post'
+				}
+			}
+		};
```
aristath: Just like in the preview script, these should be moved to a `wp_localize_script` call.

assets/js/customizer.js
```diff
@@ -0,0 +1,64 @@
...
+					// Update the value for our accessible colors for this area.
+					twentyTwentySetAccessibleColorsValue( context, to, wp.customize( 'accent_hue' ).get(), to );
+				});
+			});
+		});
+	});
+
+})( jQuery );
+
+function twentyTwentySetAccessibleColorsValue( context, backgroundColor, accentHue ) {
```
aristath: I don't like the name of this function.

assets/js/customizer.js
```diff
@@ -0,0 +1,64 @@
...
+	value[ context ] = {
+		text: colors.getTextColor(),
+		accent: colors.getAccentColor().toCSS()
+	};
+
+	// Change the value.
+	wp.customize( 'accent_accessible_colors' ).set( value );
+
+	// Small hack to save the option.
+	wp.customize( 'accent_accessible_colors' )._dirty = true;
```
aristath: It wasn't working without this line on my end, if anyone has any ideas please let me know.

ianbelanger79:
This is awesome!!! Can't wait to merge the final product

classes/class-twentytwenty-customize.php
```diff
@@ -67,11 +67,71 @@ public static function register( $wp_customize ) {
...
+					'transport'         => 'postMessage',
+					'sanitize_callback' => function( $value ) { // TODO: This should probably be a normal mehod in this object instead of a closure.
+						$value = is_array( $value ) ? $value : array();
+						foreach ( $value as $context => $values ) {
+							$value[ $context ] = array(
+								'text' => ( isset( $value[ $context ]['text'] ) ) ? $value[ $context ]['text'] : '#000000',
+								'accent' => ( isset( $value[ $context ]['accent'] ) ) ? $value[ $context ]['accent'] : '#CD2653',
+							);
+						}
+						return $value;
```
aristath: This should not be an anonymous function, it should instead be a normal method in the class. Some inline docs too wouldn't hurt.

classes/class-twentytwenty-customize.php
```diff
@@ -67,11 +67,71 @@ public static function register( $wp_customize ) {
...
+
 			/**
 			 * Colors.
 			*/
 			$accent_color_options = self::get_color_options();
 
+			/**
+			 * WIP: Disable these options 'cause we're doing the accessibility-color implementation above.
+			 */
+			$accent_color_options = array();
```
aristath: This part here disables the settings added via the `get_color_options()` method. It's just here 'cause we need to decide what to do with the previous option.

functions.php
```diff
@@ -515,3 +515,60 @@ function twentytwenty_read_more_tag() {
...
+
+	// Add script for controls.
+	wp_enqueue_script(
+		'twentytwenty-customizer',
+		get_template_directory_uri() . '/assets/js/customizer.js',
+		[ 'twentytwenty-color', 'customize-controls', 'underscore', 'jquery' ],
+		time(), // WIP: Using time() instead of a real version for cache-busting while developing.
+		false
+	);
+});
```
aristath: These will need to be moved, do we want them in the customize class? Or do we just convert them to normal functions and leave them in functions.php?

And of course we need to change the version to a normal version instead of `time()`, that's just to save some time when developing and changing scripts constantly to avoid cache issues in the browser.

functions.php
```diff
@@ -515,3 +515,60 @@ function twentytwenty_read_more_tag() {
...
+add_action( 'customize_preview_init', function() {
+	$js_dependencies = array( 'customize-preview', 'jquery' );
+	wp_enqueue_script(
+		'twentytwenty-customizer-preview',
+		get_theme_file_uri( '/assets/js/customizer-preview.js' ),
+		$js_dependencies,
+		time(), // WIP: Using time() instead of a real version for cache-busting while developing.
+		true
+	);
+} );
```
aristath: Just like the other one above, where do we want this enqueue to be in our files? And again, change `time()` to version.

aristath:
The filenames I chose will have to be changed too, I deliberately chose the wrong filenames to avoid conflicts with any other PRs that may get merged before this one.
The files should be called `customize-preview.js` and `customize-controls.js`

assets/js/color.js
```diff
@@ -0,0 +1,132 @@
...
+		];
+
+	for ( i = 0; i < rules.length; i++ ) {
+		colorFound = findColors( rules[ i ][0], rules[ i ][1] );
+		if ( colorFound ) {
+			return colorFound.color;
+		}
+	}
```
aristath: This part goes through the rules defined in the above array, and returns the first color it finds that complies with the defined ruleset.
This was the simplest solution, however I'd like to instead do this:
instead of just returning the 1st color it finds, build an array of colors that comply with the rules. They won't be many, in most cases they're no more than 5. The sort the array based on the color's saturation and return the color with the highest saturation.

classes/class-twentytwenty-customize.php
```diff
@@ -67,11 +67,71 @@ public static function register( $wp_customize ) {
...
+					'default'           => 344,
+					'type'              => 'theme_mod',
+					'sanitize_callback' => 'absint',
+					'transport'         => 'postMessage',
+				)
+			);
+
+			// Add setting to hold colors derived from the accent hue.
+			$wp_customize->add_setting(
+				'accent_accessible_colors',
```
aristath: This is a control-less setting. We're simply registering it here to make sure it gets properly sanitized with our callback. 
Since the setting doesn't have a control, we'll be manipulating its value from the JS API. It doesn't need a control, it simply needs to save the accessible colors that result from all the other controls.

pattonwebz:
This really is a very cool control to play with. We will need to come up with some good inline help text to accompany it so that users understand how it works but I like this a lot.

We can look at doing a little code cleanup and rearranging once we have had a few people test this.

Here is another gif example of it in action going from one extreme of the color selector to the other

Peek 2019-09-18 21-44

aristath:
Oh, and an additional note. If this gets merged we should change the grays. Instead of doing this:
```css
.post-meta {
    color: #6d6d6d;
}
```
we should be doing this:
```css
.post-meta {
    color: currentColor;
    opacity: .7 // or whatever value produces the same result.
}
```

That way no colors will be hardcoded and instead they'll also change along with the textcolor.

celloexpressions:
Great first pass - I'll look through the code and comment over the next day or so. For grays, we need to consider: #dcd7ca - borders and #6d6d6d - secondary text, as I noted in #284. All text, borders, icons, etc. should be updated as needed. We will likely need to reduce the relative difference between primary and secondary colors to maintain some contrast with the background color (maybe target 3:1 minimum instead of 4.5:1?).

The complete CSS "template" for colors should be built by starting with all of `style.css` and stripping it down to just the colors. This needs to happen once `style.css` is as stable as possible; all future changes to selectors with colors will also need to be synced to the JS template. 

The template is currently structured with arrays of selectors and other information. We may want to consider formatting it to look more like regular CSS for ease of updating. This could look similar to JS-templated customize controls ([example](https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentyfifteen/inc/customizer.php#L402)).

aristath:
If the array of elements is kept in PHP we only need to maintain it in 1 place. It can be passed-on the the JS file as a var

acosmin:
Looks like a winner, I'm gonna go ahead and close my PRs.

For the generated CSS, maybe you guys can do something like #304 and make them filterable.

assets/js/customize-controls.js
```diff
@@ -0,0 +1,72 @@
...
+			accent: colors.getAccentColor().toCSS()
+		};
+
+		// Change the value.
+		wp.customize( 'accent_accessible_colors' ).set( value );
+
+		// Small hack to save the option.
+		wp.customize( 'accent_accessible_colors' )._dirty = true;
+	}
+
```
celloexpressions: Let's generate the css in the customize-controls window and send it to both the server and the preview window from there. Otherwise we have to round trip to the customize preview and back.

assets/js/color.js
aristath: The input color doesn't have a saturation, it's hue only.

assets/js/color.js
aristath: > I'd rather see the desired contrast ratio as an input, with the generated color that gets returned coming as close to it as possible.

I don't get it 🤔 

> Why do we need to generate more than one accent and text color? 

Header, content footer, overlays, they all have different background-colors and so may require different accent colors.

assets/js/customize-controls.js
```diff
@@ -0,0 +1,72 @@
+/* global backgroundColors */
+/**
+ * Theme Customizer enhancements for a better user experience.
```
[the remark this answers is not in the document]
aristath: Done

assets/js/customize-controls.js
aristath: Isn't that the case with all themes and all controls though?

assets/js/color.js
```diff
@@ -0,0 +1,132 @@
+/**
```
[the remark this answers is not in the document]
aristath: ermmm for some reason when I rename the file it refuses to load in Firefox (and of course I changed it in the enqueue too).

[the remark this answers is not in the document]
aristath: We could of course just move the contents of this file in `customize-controls.js`, after all they're only used there.

assets/js/color.js
aristath: Improved this a bit

aristath:
Pushed a few tweaks a couple hours ago. Improves the selected colors, and it also picks different color for the header since it has a different background-color.

assets/js/color.js
aristath: Now sorting colors by score.
The score results as a multiplication of the contrast-with-background and the contrast-with-surrounding-text.

Also limiting the results to a minimum of 4.5:1 contrast to the background and 3:1 with surrounding text. So whatever gets picked will be OK.

classes/class-twentytwenty-customize.php
aristath: Done

functions.php
```diff
@@ -436,7 +433,13 @@ function twentytwenty_add_classic_editor_customizer_styles( $mce_init ) {
...
 
 		// Block Editor Palette.
-		$editor_color_palette = array();
+		$editor_color_palette = array(
+			array(
+				'name'  => esc_html__( 'Accent Color' ),
+				'slug'  => 'accent',
+				'color' => twentytwenty_get_color_for_area( 'content', 'accent' ),
+			),
+		);
```
celloexpressions: The editor color palette should include all 6 of the colors used in the theme that I noted in #284. The custom colors script should define the modified versions of all six colors (where black and white likely won't be modified or redefined here, but multiple accent colors could be). This is another place where we can't rely on opacity for the secondary and border colors, but need to calculate an actual color.

functions.php
```diff
@@ -523,3 +526,100 @@ function twentytwenty_read_more_tag() {
...
+					'text'   => '#000000',
+					'accent' => '#cd2653',
+				),
+			)
+		);
+
+		// If we have a value return it.
+		if ( isset( $settings[ $area ] ) && isset( $settings[ $area ][ $context ] ) ) {
+			return $settings[ $area ][ $context ];
+		}
```
celloexpressions: Let's return false or an error if there isn't a value (for example if someone uses the wrong context).

functions.php
```diff
@@ -523,3 +526,100 @@ function twentytwenty_read_more_tag() {
...
+
+if ( ! function_exists( 'twentytwenty_get_customizer_color_vars' ) ) {
+	function twentytwenty_get_customizer_color_vars() {
+		$colors = array(
+			'content' => array(
+				'setting'  => 'background_color',
+				'elements' => array(
+					'text'   => 'body',
+					// We're using #site-content a here to avoing changing styles for the header & footer in the preview.
+					'accent' => '#site-content a, .wp-block-button.is-style-outline, .has-drop-cap:not(:focus):first-letter, a.previous-post, a.next-post',
```
celloexpressions: This list is too long. Can we at least make it an array? Keep in mind that any changes to style.css now need to be synced to here. 

Ideally we would make all of these selector lists, at a minimum, more readable.

aristath:
I just went through `style.css` and compiled a list of elements with what setting should be used for their styles (apologies if JSON is not that convenient, it's just how I keep notes):
```json
{
	"content-area-accent": {
		"property: color": [
			"a",
			".color-accent",
			".color-accent-hover:hover",
			".has-accent-color",
			".has-drop-cap:not(:focus):first-letter",
			".wp-block-pullquote:before"
		],
		"property: border-color": [
				"blockquote",
				".border-color-accent",
				".border-color-accent-hover:hover"
		],
		"property: background": [
			"button",
			".button",
			".faux-button",
			".wp-block-button__link",
			".wp-block-file__button",
			"input[type='button']",
			"input[type='reset']",
			"input[type='submit']"
		],
		"background-color": [
			".bg-accent",
			".bg-accent-hover:hover",
			".has-accent-background-color",
			".comment-reply-link",
			".edit-comment-link"
		],
		"property: fill": [
			".fill-children-accent",
			".fill-children-accent *"
		]
	},
	"content-area-background-color": {
		"property: color": [
			"button",
			".button",
			".faux-button",
			".wp-block-button__link",
			".wp-block-file__button",
			"input[type='button']",
			"input[type='reset']",
			"input[type='submit']",
			".comment-reply-link",
			".edit-comment-link"
		],
		"property: background": [
			".singular .featured-media:before",
			".wp-block-pullquote:before"
		]
	},
	"header-footer-accent": {
		"property: background": [
			"social-icons a"
		],
		"property: background-color": [
			".footer-social a"
		]
	},
	"header-content-background": {
		"property: color": [
			"social-icons a",
			".overlay-header:not(.showing-menu-modal) .header-inner",
			".primary-menu ul",
			".overlay-header.showing-menu-modal .header-inner"
		],
		"property: background": [
			"#site-header",
			".menu-modal",
			".menu-modal-inner",
			".search-modal-inner",
			".archive-header",
			".singular .entry-header",
			"#site-footer"
		]
	},
	"header-content-textcolor": {
		"property: background": [
			".primary-menu ul"
		],
		"property: border-left-color": [
			".primary-menu ul ul:after"
		]
	}
}
```
The exact opacity for color `#646464` is `0.43`, so if an element has this:
```css
color: #6d6d6d;
```
was can convert to
```css
color: currentColor;
opacity: .43;
```

assets/js/color.js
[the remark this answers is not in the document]
aristath: Renamed. I don't know what was wrong yesterday, I'll blame some obscure cache.

assets/js/customize-preview.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+( function() {
+
+	// Add listener for the "header_footer_background_color" control.
+	wp.customize( 'header_footer_background_color', function( value ) {
+		value.bind( function( to ) {
+
+			// Add background color to header and footer wrappers.
+			jQuery( '#site-header,#site-footer' ).css( 'background-color', to );
```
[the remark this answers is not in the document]
aristath: Agreed. I compiled a list of elements a few hours ago and posted it on https://github.com/WordPress/twentytwenty/pull/351#issuecomment-533609825

I'll need to accomodate multiple properties, so it makes sense to me to put them all in a function that can then be added to the script via `wp_localize_script`, and then the same function can be used in `custom-css.php` to get the elements and apply the necessary CSS rules.

What do you think @celloexpressions?

classes/class-twentytwenty-customize.php
```diff
@@ -75,8 +75,85 @@ public static function register( $wp_customize ) {
...
+				array(
+					'default'           => array(
+						'content'       => array(
+							'text'   => '#000000',
+							'accent' => '#cd2653',
```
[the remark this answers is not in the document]
aristath: Hmmmm and how would we calculate the secondary color?
Just 43% between background and text? I'm saying 43 specifically because `#6d6d6d` is `hsl(0,0%,43%)` so 43% textcolor - 57% background (assuming white background & black text)

classes/class-twentytwenty-customize.php
```diff
@@ -75,8 +75,85 @@ public static function register( $wp_customize ) {
...
+			// Add the hue-only colorpicker for the accent color.
+			$wp_customize->add_control(
+				new WP_Customize_Color_Control(
+					$wp_customize,
+					'accent_hue',
+					array(
+						'label'    => esc_html__( 'Accent Color Hue', 'twentytwenty' ),
+						'section'  => 'colors',
+						'settings' => 'accent_hue',
+						'priority' => 10,
```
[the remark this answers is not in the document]
aristath: OK... It was like that for other controls in that class so thought that's what we're doing in twentytwenty, wanted things to be consistent.
I'm fine either way, I'll remove the priorities if we prefer that.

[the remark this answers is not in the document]
aristath: Removed

functions.php
aristath: I have no problem adding as many calculations and variables as we want.
However, someone will have to tell me how to get these colors.
I know how to get the color for text & accent, but what about the others? How are they calculated? Do we rotate the hue X degrees, then get a color that's 10% the rotated hue, with a saturation Y and a lightness that is 80 background - 20% text?

functions.php
aristath: Sure, np. `null` (which is what gets returned now on error) does the same job, but I can see why some people would prefer false. Making the change.

functions.php
aristath: Agreed.
Posted on the other comment above

> it makes sense to me to put them all in a function that can then be added to the script via wp_localize_script, and then the same function can be used in custom-css.php to get the elements and apply the necessary CSS rules.

Marking this one as resolved, we can discuss it in the above comment since they can be solved the same way. 👍

assets/js/customize-preview.js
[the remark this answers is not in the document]
aristath: I pushed a commit on a separate branch on my own fork just to discuss what we want to do before I push anything here.

How does this look to you? https://github.com/aristath/twentytwenty/commit/e46dc71e530babfdfe581a334943a1d4d97f0d45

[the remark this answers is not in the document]
aristath: On the other hand this is a theme, it's not a complicated framework and we're probably overthinking this.
What would be the _simplest_ solution to do?
Not the best we have in our minds, the simplest. That's what I think we should do... Otherwise this will not make it in time for the code-freeze. It doesn't need to be perfect at this point, what we want is to merge it in time so we can have another week to work-out the finer points.
It can be an array, a string, whatever as long as it works :)

aristath:
As for the cover template colors it was decided to leave them as-is for because of their complexity. There's too much involved there.. With the blend modes and the opacity controls it would not be possible to get an accessible textcolor

assets/js/color.js
aristath: Implemented the grey too.

assets/js/color.js
aristath: The implementation here was improved a lot since this comment, I think we can close this one :)

assets/js/customize-preview.js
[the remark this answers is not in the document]
aristath: Merged the arrays implementation. Added it both in PHP and JS, everything works.

classes/class-twentytwenty-customize.php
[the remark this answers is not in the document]
aristath: Now calculating and saving both the border & secondary colors.

functions.php
aristath: Added the extra colors to the palette

aristath:
@celloexpressions thank you for all the suggestions, they really made this implementation a lot better, I wouldn't have thought of some of these things without you. 🥇

ianbelanger79:
@aristath We just merged eslint into master and there are now a few errors/warning that need to be fixed, after that we will merge this

aristath:
@ianbelanger79 I was already working on fixing them. Pushed a commit just now, _should_ fix any and all issues. We'll know in a few minutes 👍

ianbelanger79:
Awesome, after travis passes, this will be getting merged.

aristath:
Yay!!!! ❤️ 🎉

ianbelanger79:
Thanks for all of your work on this @aristath and @celloexpressions. This is going to make Twenty Twenty a much better and more accessible theme.

celloexpressions:
Thanks for all of your hard work on this @aristath! This will be a great feature to use. I'll take a look through what was merged this weekend an make any comments on #284 or a new issue for any adjustments to make during beta.

PR #352, opened 2019-09-19
Another proposal for the site pagination

anace:
Related to PR #329. 

I added some changes to improve screenreader accessibility (the arrow is aria hidden and added a screenreader only "page" to every page number). Also, made the styling more consistent with the single post navigation.


carolinan:
This still affects both the single post and archive navigations?

anace:
I think so.

carolinan:
What I meant was, it should only include the changes for the site pagination. Not both :)

PR #355, opened 2019-09-19
Fixed title/logo issue with selective refresh for the Customizer

acosmin:
Regarding #335 #336

Also, made the html filterable.


acosmin:
If there's a problem with the code wrapped in `twentytwenty_site_logo()` & `twentytwenty_site_description()`, I can add the original/previous code and the selective refresh will still work.

acosmin:
@ianbelanger79 this could also be merged, it fixes the title/logo issue...

pattonwebz:
This will Fix: #335  and Close: #336

PR #356, opened 2019-09-19
JS TypeError fix when search modal is not enabled

acosmin:
If `Show search in header` is not enabled and you click on any actionable item (like the menu toggle), the console will output: `Uncaught TypeError: Cannot read property 'classList' of null at HTMLDocument.<anonymous> (VM6903 index.js:209)`

This PR will fix that, it also adds checks for the other queried Elements.


acosmin:
Fix for #357

Issue #343, opened 2019-09-18
Header jumps down in Cover Template when the modal menu is opened


ianbelanger79:
When opening the modal menu on a page with the Cover template applied, you will notice that the header will move down about 100px. See screenshots below

Menu closed
![cover-header-closed-modal](https://user-images.githubusercontent.com/34601694/65152919-dcbbfb00-d9f6-11e9-9ccf-1d8e710001ca.PNG)

Menu open
![cover-header-open-modal](https://user-images.githubusercontent.com/34601694/65152926-e04f8200-d9f6-11e9-8310-97a9788ecc86.PNG)

acosmin:
@ianbelanger79 I thinks it's from the JS part, it adds `position:fixed` on the `html` tag... why is that needed?

If you remove this line it should work:
https://github.com/WordPress/twentytwenty/blob/378becb7683d649b0dbdecbc60408de8fed020a3/assets/js/construct.js#L611

I would do a PR, but It should be fixed in #163

ianbelanger79:
This issue does not appear to have been fixed in #163

acosmin:
This can be fixed 😄 but I'm not aware of the implications, @andersnoren does removing position:fixded have an effect on the modals?

andersnoren:
@acosmin Nope, the old contents of the `scollLock` function can be removed altogether and replaced with a scroll to the top and setting the `body to `overflow: hidden`. That will fix the issues we have with the admin bar as well.

The old `scrollLock` (in Chaplin) needed to retain the scroll position due to a sticky header option being included, but that's not needed in Twenty Twenty since the toggles are always at the top of the screen.

andersnoren:
Fixed by #360.

PR #360, opened 2019-09-19
Removed `twentytwenty.scrollLock` and fixed `overflow` issue

acosmin:
Fix for #343

@andersnoren is something like this what you wanted? It fixes the main issue and removes/adds the overflow on body when needed.

I didn't really get the scroll to top part 😄 hope it's what you wanted.


acosmin:
removed the mentioned function

PR #361, opened 2019-09-19
Changed cursor for close state in modals

acosmin:
😄 If this is even wanted. It's a png, couldn't get svg to work 😞


acosmin:
@andersnoren is this something that is wanted :) or should I close it?

andersnoren:
@acosmin Sorry, meant to get back to you on this one but it slipped my mind :( I think it's good to have a cursor that indicates that it can be closed with a click, but since Edge and IE doesn't support base64 values for the `cursor` property, I think it's better if we use one of the default cursors.

Not sure which one is the most appropriate though. There doesn't seem to be a close/cross equivalent. An arrow pointing left or right feels like it could be confusing, so maybe the `pointer` one is the safest choice.

acosmin:
@andersnoren that's a good idea, changed it to pointer. Also, we could probably add a `title` attribute on `div.menu-modal` and `div.search-modal`, something like `Close Modal`...

andersnoren:
@acosmin Sorry for being late with a reply on this one (again). I discovered that there's an issue in IE11 and `cursor: default`. Instead of the cursor being set to its default state, it's becomes set to inherit the cursor of the parent, so currently, all children of the `.cover-modal` element get the `pointer` cursor in IE11. Changing it to:

```
.cover-modal.show-modal > * {
	cursor: default;
}
```

...gives us the same behavior in modern browsers, and fixes the issue in IE11. I promise that it won't take me six days to get back to you the next time :)

Issue #35, opened 2019-09-07
Add build process for linting and code optimization


mor10:
Related: #19 and #34 

Consider adding a build process with Babel for JS and [PostCSS](https://www.npmjs.com/package/postcss) and [`postcss-preset-env`](https://www.npmjs.com/package/postcss-preset-env) for CSS to streamline and simplify builds. PostCSS includes Autoprefixer. The build process can also automate code linting.

The build process can either be straight-up NPM scripts or a basic Gulp process. I do not recommend Webpack for this as it is primarily for packaging JavaScript bundles which isn't applicable in this context. 

cc @ataylorme

senadir:
why do we need a js build process? also, we can use @wordpress/scripts to handle all of those things to us behind the scenes and be actively developed outside the theme

dingo-d:
> The build process can either be straight-up NPM scripts or a basic Gulp process. I do not recommend Webpack for this as it is primarily for packaging JavaScript bundles which isn't applicable in this context.

While Webpack is a bundler primarily, it's usage is exactly to use things like PostCSS and other tools that can be used for building JS and CSS. Since core uses Webpack, it would be beneficial to use it for the build process, just to get people familiar with it.

ataylorme:
I would like to advocate for using more modern CSS, like custom properties, instead of Sass. Nothing against Sass but I believe the core themes should show what is capable with the CSS spec. Modern browsers support quite a bit and PostCSS can be used to provide support for legacy browsers.

Regarding PostCSS and Webpack, I don’t think it has to be one or the other. There is [a PostCSS loader for Webpack](https://github.com/postcss/postcss-loader).

mor10:
Absolutely for custom properties and modern CSS over Sass. As you well know.

dingo-d:
Can we use custom CSS properties? https://caniuse.com/#feat=css-variables
IE is not supporting them (surprise, surprise! xD)

ataylorme:
> Can we use custom CSS properties? https://caniuse.com/#feat=css-variables
IE is not supporting them (surprise, surprise! xD)

That is _exactly_ what PostCSS is for.

> PostCSS Preset Env lets you convert modern CSS into something most browsers can understand, determining the polyfills you need based on your targeted browsers or runtime environments.

[See the PostCSS preset](https://github.com/csstools/postcss-preset-env#postcss-preset-env-) for an example of conversions done.

[...]

PR #363, opened 2019-09-19
add @wordpress/scripts package

Ismail-elkorchi:
Related : https://github.com/WordPress/twentytwenty/issues/35

This PR adds `@wordpress/scripts` package, and adds the `lint:css` and `lint:js` scripts to `package.json`. It also adds Travis configuration to make use of the new scripts.


ntwb:
Looking great @Ismail-elkorchi 

I've added a `.stylelintrc.json` file to exclude the `no-descending-specificity` rule (this is also done in Gutenberg)

I'll make a couple more tweaks later as required

ntwb:
I've updated all the files that will be covered by the _lint_ additions in this PR.

We should merge those first I believe to help prevent Travis CI failing on all pull requests until the changes are made.

ntwb:
There are two issues, #408 and #409 that require resolving for the stylelint lint to tests to pass, both these should be resolved before merging this PR.

ntwb:
There is one issue, #410 that requires resolving for the ESLint lint to tests to pass, this should be resolved before merging this PR.

ianbelanger79:
This PR is failing Travis because of the ES5 JS we are using. Is there a way to tell it to use ES5 standards? @Ismail-elkorchi

Ismail-elkorchi:
@ianbelanger79  I've added a `.eslintrc`file to make use of the ES5 rulesets.

But, as I said in https://github.com/WordPress/twentytwenty/issues/410#issuecomment-533799823, I think that writing code in ES6 syntax and transform it using the build script in `@wordpress/scripts` is best approach to take.

Ismail-elkorchi:
I've added a rule in the `.eslintrc` file to use only the browser environment. This will prevent Travis from complaining about undefined objets when it's irrelevant.

Ismail-elkorchi:
@ianbelanger79 I'll make a PR to fix the remaining errors in the `index.js` file.

ianbelanger79:
I think @acosmin is working on that, but if you can get it done first, go ahead, we are running out of time quickly

acosmin:
I'm done, wait 2 minutes :)

Issue #366, opened 2019-09-19
Page jumps left/right when opening the search modal on desktop


nielslange:
**Steps to reproduce**

1. Open a long page, e.g. a blog posts page
2. Open the search modal

**Expected behaviour**

The underlying page remains on the same spot.

**Current behaviour**

The underlying page moves to the right when the search modal gets opened and back to the left when the search modal gets closed.

https://cld.wthms.co/i3nfbR

It seems I can tackle this issue by adding `overflow: overlay;` to the `body` element. I'll give it a try to see if my assumption works out. 😀

acosmin:
Might be related to this PR and its issue https://github.com/WordPress/twentytwenty/pull/360

ianbelanger79:
This happens when you open the menu modal as well. It's due to the scrollbar being removed when a modal is open. `overflow: overlay;` on the body does not work because opening the modal adds `overflow: hidden;` to the body as inline style

nielslange:
@acosmin Yes, that sounds related to #360 but the page still jumps despite your PR had been merged.

@ianbelanger79 On my end the page does not jump any longer when I add `overflow: overlay;` to the `body` element as that reduces the `body` width resp. ensures that the scrollable sidebar does not push the `body` to the left. Shall I create a quick PR so that you can check out what I mean?

joyously:
[MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow) has thumbs-down for `overlay` 
> Only supported in WebKit-based (e.g., Safari) and Blink-based (e.g., Chrome or Opera) browsers.

nielslange:
@joyously Absolutely. I just saw https://developer.mozilla.org/en-US/docs/Web/CSS/overflow. `overflow: overlay;` seems to be the wrong decision to fix this issue.

acosmin:
tried something, needs testing :)

nielslange:
@acosmin Did you tried it in #360?

nielslange:
While I don't like `!important` statements, adding `overflow: auto !important` to the `body` element should also fix the issue.

acosmin:
@nielslange we've added `overflow: hidden` with the purpose of not allowing scrolling 😄 that's against it...

@andersnoren any thoughts on this?

nielslange:
@acosmin Is there any other approach to avoid scrolling without having the page jumping if the modal opens?

acosmin:
[...]

PR #367, opened 2019-09-19
UPDATE #366: Page jumps left/right when opening the search modal on desktop

acosmin:
needs testing: #366


acosmin:
I think it's there 👍 please test it.

Issue #353, opened 2019-09-19
Horizontal menu submenu items are not working with keyboard only


carolinan:
For this menu, the submenu items are visible when you focus on the parent menu item. But when you press tab again to access the submenu item, it is no longer visible.

PR #368, opened 2019-09-19
Fix keyboard navigation issue in Primary Menu.

anace:
This PR solves issue #353 

Keyboard navigation for the primary menu (the horizontal menu in desktop) is not working well. When navigating with the keyboard, the submenus are visible when the focus is on the parent element, but are not visible when the focus is on the children elements. A solution to this issue is

1. to add the class `focus` to the parent element when the focus is on a child element (i.e. when the focus is on the `a` element, the `li` element that contains it get the class `focus`).

2. to style the `focus` class as the `:hover` and `:focus` pseudo-classes.

I added a function at the end of `index.js` to do this. However, it may be better to integrate this functionality in Focus Management. Also, I added two lines in `style.css`.

WordPress username: acalfieri


joyously:
Did you use `focus-within`?

anace:
No. I added the class focus with JS.

carolinan:
This PR  must not contain any changes to the post navigation, it should only contain code related to the menu.

anace:
I think you can choose the lines that you will accept.

anace:
This should be OK when the checks are completed.

anace:
This is not working well. I'll make another one.

PR #374, opened 2019-09-20
Update `.editorconfig` to latest WordPress version

ntwb:
This updates the `.editorconfig` to the latest WordPress version:

https://core.trac.wordpress.org/browser/trunk/.editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -19,3 +19,6 @@ indent_size = 2
 
 [*.md]
 trim_trailing_whitespace = false
+
+[*.txt]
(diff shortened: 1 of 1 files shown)


pattonwebz:
I added a line to the config to set *.txt files to use `crlf` as line endings.

pattonwebz:
@nielslange I added a block for the line endings. I'm gonna merge this now but tell me if you think we should adjust it a little more :)

ntwb:
Good catch @nielslange, I forgot about that use case, thanks

Issue #371, opened 2019-09-20
big, small have font-style instead of font-size


joyously:
This was added in #259, but it is incorrect.
https://github.com/WordPress/twentytwenty/blob/8e28522319df7bd45433e8a4f04438722e7d4786/style.css#L388-L389

It needs to be font-size.

PR #378, opened 2019-09-20
Updated big and small element styles in style.css from font-style to …

Alloffices:
updated big and small element set to font-size


andersnoren:
@Alloffices This was fixed in #371, so I'm closing this one.

Issue #320, opened 2019-09-17
table style : editor versus front end


joyously:
Tables are styled differently in the classic editor stylesheet than in the block editor and the front end.
Which one is matching the design?

Classic editor: (this is a table, not a table block, but it looks the same as a table block)
![classic-editor-table-style](https://user-images.githubusercontent.com/7957680/65005972-52c14480-d8c7-11e9-995e-cb4081a05165.jpg)

Block editor: (this is a table block)
![block-editor-table-style](https://user-images.githubusercontent.com/7957680/65005970-5228ae00-d8c7-11e9-9b7a-d39d29e97870.jpg)

Front end: (this looked the same whether it was a block or not)
![front-end-table-style](https://user-images.githubusercontent.com/7957680/65005971-52c14480-d8c7-11e9-962a-e71fe4ded3b1.jpg)

joyously:
With the latest changes, the Classic Editor now matches the front end. Block editor remains the same as before.
![Edit-table-classic-editor](https://user-images.githubusercontent.com/7957680/65376668-3b3be000-dc68-11e9-9dde-f252f3f6bf4f.jpg)

ianbelanger79:
Possible fix for this #699

joyously:
Mostly the difference is `table-layout: fixed` and the borders, not padding.

Issue #117, opened 2019-09-08
Using Inter font family in editor style


vbaimas:
I believe that one of nicest things that we can do for your users is to implement a stylesheet specifically for the editor so that they get a true WYSIWYG experience.

So, I like the idea of having the Inter font itself apply to the editor style.

carolinan:
This has been fixed, so I am closing the issue. Thank you!

PR #384, opened 2019-09-20
Update block editor styles to match the front-end

### LINKED ISSUE #287: code block editor styling
the code block has an additional white background in the editor, that is not there on the frontend. There probably should be consistency between the two :)

Also is the background should stay it should get some additional padding.

Screenshot 2019-09-15 at 19 51 09

andersnoren:
Updates the block editor styles to match the front-end output of Twenty Twenty. Tested with Gutenberg 6.4.0, Gutenberg 6.5.0 and with the plugin deactivated (on 5.2.3).

Fixes #287 
Partial fix for #320 and #117.


andersnoren:
I've updated the styles to try and make them as close to the standards as possible, but another linting will be needed when #363 is merged.

PR #392, opened 2019-09-20
#371 Updated : big, small have font-style instead of font-size

tm-const:
changed font-style to font-size to both big and small css selectors


ianbelanger79:
This was already fixed in a previous PR

PR #393, opened 2019-09-20
Removed Outline from input and buttons

tm-const:
Removed outlines from buttons and input fields

**from:** 

Screen Shot 2019-09-20 at 8 03 30 AM

**to:**

Screen Shot 2019-09-20 at 8 03 43 AM

### DIFF
--- a/style.css
+++ b/style.css
@@ -714,6 +714,7 @@ textarea {
 	max-width: 100%;
 	padding: 1.5rem 1.8rem;
 	width: 100%;
+	outline: none;
 }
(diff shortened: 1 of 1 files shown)


acosmin:
Those have a purpose for a11y :) you shouldn't do that

carolinan:
Yes the outline needs to be kept, please see 
https://make.wordpress.org/core/handbook/best-practices/coding-standards/accessibility-coding-standards/#keyboard-accessibility

Issue #390, opened 2019-09-20
Change responsive header layout when search is hidden


yannickiki:
Currently if we choose to hide the search in the header, the Logo/Site Title is centered with the menu icon at the right.

Maybe it would be better to put the logo to the left.  @andersnoren What do you think?

![menu](https://user-images.githubusercontent.com/54663751/65319388-d3ba5d80-db9f-11e9-8e17-b7ae6f9f3c09.png)

nielslange:
Personally, I prefer the centred title even if the search icon is hidden on mobile.

andersnoren:
@yannickiki Sorry for the late reply. Agreed. It looks odd compared to the default structure of Twenty Twenty, but visitors won't have that prior knowledge and will except it to be displayed to the left on mobile.

andersnoren:
Fixed by #404.

PR #404, opened 2019-09-20
Change responsive header layout when search is hidden

acosmin:
Fix for #390


acosmin:
@andersnoren sure thing 👍

PR #406, opened 2019-09-20
WPCS : Remove twentytwentyGetQueryStringValue() 

Ismail-elkorchi:
While looking at https://github.com/WordPress/twentytwenty/pull/363, I found that the Travis CI complain about this function being unused. I did a manual search, and confirmed that it is not used anywhere.


ianbelanger79:
Already removed when #422 was merged

PR #419, opened 2019-09-21
Add screen reader text and filter to entry-header categories

### LINKED ISSUE #412: Category section is missing a screen reader text
The category section does not have a screen reader text, so the context for these links is missing.

andersnoren:
Added "Categories" as screen reader text before the output of categories in the entry header, and added a filter allowing child themes to disable the output of categories in this location altogether.

Fixes #412.


carolinan:
The cover template does not use entry header, these changes also needs to be implemented here:
https://github.com/WordPress/twentytwenty/blob/master/template-parts/content-cover.php#L62

andersnoren:
@carolinan Oh right. Updated!

template-parts/entry-header.php
```diff
@@ -15,10 +15,14 @@
 
 		<?php
 
-		if ( has_category() ) {
+		// Allow child themes and plugins to filter the display of the categories in the entry header.
+		$show_categories = apply_filters( 'twentytwenty_show_categories_in_entry_header', true );
```
pattonwebz: Could we use the special return functions here instead of static bools? https://developer.wordpress.org/reference/functions/__return_true/

Issue #410, opened 2019-09-21
Fix outstanding WordPress JavaScript Coding Standards


ntwb:
As part of #363 for the ESLInt lint tests to pass the following issues require resolving:

```shell
/home/travis/build/WordPress/twentytwenty/assets/js/index.js

    5:1   error  Unexpected var, use let or const instead                     no-var

    9:27  error  'NodeList' is not defined                                    no-undef

   10:2   error  'NodeList' is not defined                                    no-undef

   23:15  error  'Event' is not defined                                       no-undef

  103:8   error  'adminBar' is never reassigned. Use 'const' instead          prefer-const

  104:5   error  'newHeight' is defined but never used                        no-unused-vars

  193:2   error  Expected method shorthand                                    object-shorthand

  193:21  error  Unexpected space before function parentheses                 space-before-function-paren

  194:49  error  Unexpected space before function parentheses                 space-before-function-paren

  195:4   error  Unexpected var, use let or const instead                     no-var

  196:4   error  Unexpected var, use let or const instead                     no-var

  197:4   error  Unexpected var, use let or const instead                     no-var

  198:4   error  Unexpected var, use let or const instead                     no-var

  253:50  error  Unexpected mix of '*' and '+'                                no-mixed-operators

  253:58  error  Unexpected mix of '*' and '+'                                no-mixed-operators

  452:7   error  Unexpected var, use let or const instead                     no-var

  554:10  error  'twentytwentyGetQueryStringValue' is defined but never used  no-unused-vars

  671:26  error  'HTMLElement' is not defined                                 no-undef

  690:21  error  Unexpected mix of '*' and '+'                                no-mixed-operators

  690:25  error  Unexpected mix of '*' and '+'                                no-mixed-operators

  693:17  error  Unexpected mix of '*' and '+'                                no-mixed-operators

  693:23  error  Unexpected mix of '*' and '-'                                no-mixed-operators

  693:35  error  Unexpected mix of '*' and '-'                                no-mixed-operators

  693:41  error  Unexpected mix of '*' and '+'                                no-mixed-operators
```

To test `index.js` with ESLint checkout the #363 PR and run the following:

> `npm run lint:js`

cc: @acosmin @fabiankaegy could you both take a look at fixing these issues as you're both familiar with the code here from your work in #163 please 🙇🏼‍♂️

acosmin:
@ntwb did it, but this will not work for IE11 (older browsers) if you don't build it for es5

[...]

PR #422, opened 2019-09-21
index.js - back to es5

acosmin:
- Removed dependencies
- Converted back to es5
- Removed `twentytwentyGetQueryStringValue()`, not used.

For #410 

@ntwb You should probably merge this before adding linting and we can fix any issues if it reports something.


ianbelanger79:
@acosmin Travis is failing when I try to merge #363 and it is due to numerous errors in the index.js file, see travis results here https://travis-ci.com/WordPress/twentytwenty/jobs/237803811

acosmin:
@ianbelanger79 I can fix those errors, just one thing, can I please add `/*eslint-env browser*/` https://eslint.org/docs/rules/no-undef#environments

This is for all the `window, NodeList is not defined`... they are defined 😄

ianbelanger79:
Absolutely add `/*eslint-env browser*/` if needed.

Ismail-elkorchi:
@ianbelanger79 @acosmin I'll make Eslint use the browser environment in https://github.com/WordPress/twentytwenty/pull/363. This will get rid of  errors about the undefined variables.

PR #426, opened 2019-09-21
#354 - Fix modal menu keyboard tab navigation to close button at the end

### LINKED ISSUE #354: Closing the expanding menu with keyboard only requires tabbing backwards
When you reach the last menu item of the open, expanded menu (in this case a social link), the next item when you tab is on the page. It is difficult the see where the focus is, because the menu remains open. 

Expected: The next item should be the close button.

Currently, to reach the close button we need to tab backwards through all the menu items.

c4rl0sbr4v0:
This pull requests fixes #354 
Tested on Chrome and Safari on Mac:

Without social links on the menu:
menu_navigation_tabs_no_social

With social links on the menu:
menu_navigation_tabs


c4rl0sbr4v0:
I close it, It still needs some testing

PR #432, opened 2019-09-21
fix stylelint errors

ianbelanger79:


ntwb:
I'll get stylelint upstream update to fix the dual values for `font-weight`, seems like a bug that should not require a _disable_

PR #439, opened 2019-09-21
Use core's function for navigation between single posts.

anace:
This PR solves issue #300, as in PR #314 and PR #350 . The markup is simpler and accessibility ready.


swissspidy:
Can you undo all the unrelated file permission changes?

anace:
I deleted that fork in the meantime. I may close this PR and make a new one.

PR #443, opened 2019-09-21
Improve the way customizer styles get updated

aristath:
See #284 
This PR changes the way styles get applied to the preview pane.

Instead of them getting applied directly to the elements themselves, they will now be added to a stylesheet in `<head>`.
This will help us avoid issues and hacks like this one: https://github.com/WordPress/twentytwenty/blob/049e8fe153aacdadbb14ed83030a08ec8a5c782a/functions.php#L701-L702
Styles will be more consistent and in line with the normal frontend.


joyously:
Yes, this is a better way. 
Do you think it handles the case of the preview refreshing due to changing a different option?
I saw a glitch while testing the colors earlier, but then I couldn't recreate it. Please try a scenario like
- mess around with background color and accent color
- go to Theme Options and change from full text to summary

(what I saw once only: when changing to summary, the page refreshed and the header was black)

aristath:
>  Do you think it handles the case of the preview refreshing due to changing a different option? I saw a glitch while testing the colors earlier, but then I couldn't recreate it.

Good catch! 👍 
Added fix on https://github.com/WordPress/twentytwenty/pull/446

PR #445, opened 2019-09-21
#354 - Fix tab navigation

### LINKED ISSUE #354: Closing the expanding menu with keyboard only requires tabbing backwards

c4rl0sbr4v0:
Return to close button after tabbing to the last item of the menu.

This pull requests fixes #354 
Tested on Chrome and Safari on Mac:

Without social links on the menu:
menu_navigation_tabs_no_social

With social links on the menu:
menu_navigation_tabs

With two menus:
menu_navigation_tabs_no_social_two_menus


carolinan:
Hi,
please try to fix the error from the Travis report.

The focus needs to be contained within the modal, it is still possible to tab backwards (shift + tab) 
and focus on items on the page hidden the menu.

When I tab away forward from the last menu item, the first menu item is focused, and not the close button.

c4rl0sbr4v0:
Hi! I will work on that tomorrow morning.
I put the prevent default in order not to happen the last issue "When I tab away forward from the last menu item, the first menu item is focused, and not the close button." 
Can you please tell me with how that happened (Browser and system)

Thanks!

carolinan:
I use Chrome Version 77.0.3865.90 on windows.
I don't know JS well enough to help you further.  This codepen was recommended, if it helps: 
https://dev.to/islam/how-to-trap-focus-inside-a-modal-2791

We have a commit freeze for beta 1 in a few hours :)

c4rl0sbr4v0:
Let's fix it before then :-)

carolinan:
This works for the expanded menu in Chrome, Firefox and Edge but not in IE11.

For the mobile menu, there is no close button, so we need to pick a focus target for when we tab forward from the last item.
This should probably be the menu toggle,  mobile-nav-toggle,
or, it could be the search if the search is enabled, or the site title, since both these items comes _before_ the menu toggle in the visual order.

c4rl0sbr4v0:
Cool! I will try it in IE asap :-) Also I will update the mobile part :-)

carolinan:
@acosmin Do you have time to look at this? :)

acosmin:
@carolinan Let's wait for #557 to be merged and after that the conflicts should be fixed. Maybe @sgomes could also take a look at this.

c4rl0sbr4v0:
I will test it today on IE, conflicts for my part are easy, I just add a function an execute it :-)

c4rl0sbr4v0:
@carolinan Updated and tested on IE11, please let me know if is also working for you. 

> For the mobile menu, there is no close button, so we need to pick a focus target for when we tab forward from the last item.
> This should probably be the menu toggle, mobile-nav-toggle,
> or, it could be the search if the search is enabled, or the site title, since both these items comes _before_ the menu toggle in the visual order.

This will be fixed on Friday by 13:00 UTC+2 :-)

c4rl0sbr4v0:
Everything working as requested!

carolinan:
Mobile menu:
+ When I tab forward from the last menu item, it now targets the menu toggle.
- I can't tab backwards from the menu toggle. It does not matter if the menu is open or not.

carolinan:
OK after testing and discussing we will merge this now and work from here to do some changes to the mobile menu, including adding the close button.

PR #449, opened 2019-09-21
Update post edit link to match entry-content styles

andersnoren:
Updates the "Edit This" link to have the same font and font size as the rest of the entry content.


joyously:
My opinion: the Edit link should stand out, not blend in to the content...

I was testing earlier and wondered if it is a good thing to have the Edit link `display: block`, because the cursor changes when hovering anywhere along the entire width, and inadvertently clicking on that white space will navigate to the link. That can be very annoying when switching windows by clicking white space.

carolinan:
What about edit-comment-link? :) Right now the Edit comment and Reply to comment look identical.

andersnoren:
Fair point, @joyously. I've updated the PR to make the edit post link match the style and structure of the post meta items, to make it stand out more from the entry content.

image

PR #451, opened 2019-09-22
Add a searchform.php template

pattonwebz:
Adds a first searchform.php template that gets used in the 404 template and the 'no posts found' search template. Labels and `aria-*` could use improvements.

The modal search form is unchanged by this PR. The styles on the other search also is unchanged. I've attached screenshots of the other searches in a widget, the no post found block and the 404 search block.

Screenshot from 2019-09-22 01-34-48
Screenshot from 2019-09-22 01-34-52
Screenshot from 2019-09-22 01-35-38

### DIFF
--- a/searchform.php
+++ b/searchform.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * The searchform.php template.
+ *
+ * Used any time that get_serach_form() is called.
+ *
+ * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
+ *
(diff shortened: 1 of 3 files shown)


searchform.php
```diff
@@ -0,0 +1,27 @@
+<?php
+/**
+ * The searchform.php template.
+ *
+ * Used any time that get_serach_form() is called.
```
GlennMartin1: `get_serach_form()` should be `get_search_form()`

Issue #118, opened 2019-09-08
Support for non-latin font fallbacks


andersnoren:
Following in the tradition of [this issue](https://github.com/WordPress/twentynineteen/issues/29) for Twenty Nineteen and [this one](https://github.com/WordPress/twentyseventeen/issues/59) for Twenty Seventeen, we’d love to get some help from all of the polyglots and i18n folks out there to make sure that Twenty Twenty has awesome support for non-latin alphabets. To start with, we need help with finding good non-latin fallbacks. Later on, we'll need lots and lots of help with testing. 

Kudos to @nielslange for this handy table from last year, which I'm stealing outright. This is the list of alphabets supported in `twentyseventeen` and `twentynineteen`. I'll update with font stacks, testers and PRs for Twenty Twenty as we go.

| Language | ISO Codes | Fonts | Tester | Status |
| :--- |:--- |:--- |:--- |:--- |
| Arabic | ar, ary, azb, ckb, fa-IR, haz, ps | Tahoma, Arial, sans_serif; | @elazzabi | [✅](https://github.com/WordPress/twentytwenty/issues/118#issuecomment-536925675) |
| Chinese Simplified (China) | zh_CN | "PingFang SC", "Helvetica Neue", "Microsoft YaHei New", "STHeiti Light", sans-serif   | @yuliyang | — |
| Chinese Traditional (Taiwan) | zh_TW | "PingFang TC", "Helvetica Neue", "Microsoft YaHei New", "STHeiti Light", sans-serif;  | @yuliyang | — |
| Chinese (Hong Kong) | zh_HK | "PingFang HK", "Helvetica Neue", "Microsoft YaHei New", "STHeiti Light", sans-serif;  | @yuliyang | — |
| Cyrillic | bel, bg-BG, kk, mk_MK, mn, ru-RU, sah, sr_RS, tt_RU, uk | "Helvetica Neue", Helvetica, "Segoe UI", Arial, sans_serif; | — | — |
| Devanagari | bn_BD, hi_IN, mr, ne_NP | Arial, sans_serif; | — | — |
| Greek | el | "Helvetica Neue", Helvetica, Arial, sans_serif; | — | — |
| Gujarati | gu | Arial, sans_serif; | — | — |
| Hebrew | he-IL | "Arial Hebrew", Arial, sans_serif; | — | — |
| Japanese | ja | sans-serif; | @nukaga, @miminari | — |
[...]

PR #457, opened 2019-09-22
UPDATE #455: Add instructions for build tools to CONTRIBUTING.md

### LINKED ISSUE #455: Add instructions for build tools to CONTRIBUTING.md
As we've introduced build tools recently, the steps to run them should be mentioned in CONTRIBUTING.md. 

I'm working on this issue and will create a PR within the next minutes.

nielslange:
Fixes #455 

Added the following section:

## Create Right-To-Left (RTL) CSS file

Twenty Twenty uses [RTLCSS](https://rtlcss.com/) which allows to transform the  Left-To-Right (LTR) file `style.css` into the Right-To-Left (RTL) file `style-rtl.css`.

### Installation instructions

1. Using a command line interface, go to the `twentytwenty` directory `cd /my-compter/local-wordpress-install/wp-content/themes/twentytwenty`.
2. Type `npm install` into the command line and press [return], to install all Node.js dependencies. 
3. The dependencies may take a few minutes to download and once it completes, you’re done.

### Usage instructions

1. After making a change to the `style.css` file, which affects RTL, run `npm run build:rtl` from within the theme directory to build the `style-rtl.css` file with your new changes.


joyously:
@nielslange 
There was an open question in one of the RTL issues about whether there are RTL versions of the editor stylesheets.

nielslange:
@joyously I must have missed that and, in my opinion, the editor stylsheets should also be converted to ensure that they show the correct direction. I'll analyse this issue and enhance the PR if needed.

package.json
```diff
@@ -30,6 +30,8 @@
   },
   "scripts": {
     "build:rtl": "rtlcss style.css style-rtl.css",
+    "build:rtl-esb": "rtlcss assets/css/editor-style-block.css assets/css/editor-style-block-rtl.css",
```
nielslange: @ntwb Would this be the correct way to create the RTL file for `assets/css/editor-style-block.css`?

package.json
```diff
@@ -30,6 +30,8 @@
...
+    "build:rtl-esc": "rtlcss assets/css/editor-style-classic.css assets/css/editor-style-classic-rtl.css",
```
nielslange: @ntwb Would this be the correct way to create the RTL file for `assets/css/editor-style-classic.css`?

nielslange:
@joyously As you can tell, I've enhanced the PR for this issue as the editor stylesheet files contained a few align statements. Thanks for pointing out this issue.

joyously:
@nielslange  Does that include getting the new ones loaded?

nielslange:
> @nielslange Does that include getting the new ones loaded?
That would be good good to be true, wouldn't it! I totally missed enqueueing them. 🤦‍♂️

joyously:
Well, the add_editor_stylesheet function does allow for RTL. I just wanted to make sure that's what was happening.

nielslange:
@joyously I simply forgot to enqueue the RTL editor stylesheets initially. Also, if I understand https://developer.wordpress.org/reference/functions/add_editor_style/ correctly, then I don't need to enqueue `/assets/css/editor-style-classic-rtl.css` but only `/assets/css/editor-style-block-rtl.css`. Can you confirm that?

joyously:
The function only adds the RTL for the first stylesheet passed. If you are loading the block editor sheet differently, then you will have to do your own load for RTL.

nielslange:
In this case, it should work already. But I’ll keep it in the mind when coordinating the testing for https://github.com/WordPress/twentytwenty/issues/118.

package.json
ntwb: Close, I’ll add some functionality so that there is only a single bpm script to run instead of 3

ntwb:
I've update the scripts so that the single npm script `npm run build:rtl` builds all 3 RTl _concurrently_ (in parallel as a single task)

nielslange:
> I've update the scripts so that the single npm script `npm run build:rtl` builds all 3 RTl _concurrently_ (in parallel as a single task)

Thanks, @ntwb! Your solution is much more elegant. I'll keep that solution in mind for related tasks in the future. 😀

PR #465, opened 2019-09-22
UPDATE #462: Add role to header and aria-label to sidebars

### LINKED ISSUE #462: Add missing aria roles
The site header is missing the aria landmark role.

For reference: https://make.wordpress.org/themes/handbook/review/accessibility/required/#aria-landmark-roles

nielslange:
Fixes #462 

@carolinan In addition to adding `role="banner"` to `header.php` I've also added `aria-label="<?php esc_attr_e( 'Footer #1', 'twentytwenty' ); ?>"` and `aria-label="<?php esc_attr_e( 'Footer #2', 'twentytwenty' ); ?>"` to `footer.php`.


pattonwebz:
The extra labels on the footer widget sections for clarity I think is a great idea. In researching this just now I found out that the `<header>` doesn't actually need the `role="banner"` since this is the only header element and is a direct decedent of the body tag, I left some more details in this comment: https://github.com/WordPress/twentytwenty/issues/462#issuecomment-533917235

carolinan:
I am not sure the # translates so well here. It should probably be removed in other places too.

carolinan:
Overall I think there should only be one widget are _displayed_ in columns 😉

nielslange:
@pattonwebz I'm not sure I understand what to do exactly. You mentioned in https://github.com/WordPress/twentytwenty/pull/465/#issuecomment-533917373 to remove `role="banner"` from the `<header>` but according to @afercia 
 in https://github.com/WordPress/twentytwenty/issues/462#issuecomment-533982226 the `<header>` should _"use an explicit ARIA role in addition to the HTML5 element"_. Shall I remove `role="banner"` or keep it?

@carolinan Can you give me more context on your two comments as well? I'm not really sure what to do here.

afercia:
Thanks for the ping :)

An `aria-label` on a `div` element without any ARIA role won't have any effect. These are just divs that define the 2 "footer" widget areas.

Instead, the real footer, in terms of HTML5 and semantics, it's the `<footer>` element.
`<footer id="site-footer" role="contentinfo"  ...`

There must be only one `<footer>` in the scope of the body (ie. direct child of body). It already has a `role="contentinfo"`. Since it's only one it doesn't need an `aria-label` to be distinguished from other footers.

In short, I'd recommend to not add the aria-labels.

## Semantics

Testing the theme merged in core, I see the current `<footer>` element wraps:
- the footer navigations
- the widget areas
- the copyright /powered by / to the top area

Screenshot: see the highlighted area

Screenshot 2019-09-24 at 16 06 45

I'm not sure this is ideal from a semantic perspective. A body `<footer>` should only contain information about the parent document:

https://www.w3.org/TR/wai-aria-1.1/#contentinfo
> Examples of information included in this region of the page are copyrights and links to privacy statements.

I'd recommend to refactor in order to have:
- the navigation menu should be within a `<nav>` element with `role="navigation"` 
- the widget areas should be within an `<aside>` element with `role="complementary"`
- the `<footer>` with `role="contentinfo"` should contain only the copyright /powered by / to the top

pattonwebz:
@nielslange sorry for the confusion. I was mistaken and I edited my comment above. Follow the other advice here and not mine :)

nielslange:
> I'd recommend to refactor in order to have:
> - the navigation menu should be within a <nav> element with role="navigation"
> - the widget areas should be within an <aside> element with role="complementary"
> - the <footer> with role="contentinfo" should contain only the copyright /powered by / to the top

@afercia Thanks for your detailed reply. I'll create a separate issue for restructuring the footer and will implement your other feedback in this issue.

> @nielslange sorry for the confusion. I was mistaken and I edited my comment above. Follow the other advice here and not mine :)

@pattonwebz Don't worry! I'll follow @afercia's advice.

andersnoren:
@nielslange There doesn't seem to be any changes in this PR at this point. Can you take a look?

nielslange:
@andersnoren Sure, I'll work on the PR of this issue and then take care of the PR for https://github.com/WordPress/twentytwenty/issues/564.

ianbelanger79:
Closing because there are no updates to this PR

PR #466, opened 2019-09-22
UPDATE #317: Exclude alignleft and alignright from RTLCSS

### LINKED ISSUE #317: alignleft and alignright are reversed in RTL
The alignleft class should always be on the left, and alignright should always be on the right, even in RTL languages.
alignments-reversed

nielslange:
Fixes #317


joyously:
Looks like this needs a refresh, but it doesn't address the editor stylesheets either.

nielslange:
@joyously I'll look up this issue now.

PR #467, opened 2019-09-22
Rename license.txt to LICENSE

nielslange:
Renamed license file name to reflect the used license within the summary of https://github.com/WordPress/twentytwenty:

Screenshot 2019-09-22 at 15 35 05


jeffpaul:
@nielslange @ntwb is there a .org requirement for themes to have a license.txt file or are we ok to make this change?

ntwb:
@jeffpaul Themes should have a license, and this change matches the 2019 format:
https://github.com/WordPress/twentynineteen/blob/master/LICENSE

joyously:
It's not a required file, but it's recommended. The name doesn't matter.

nielslange:
@jeffpaul & @joyously By renaming the file the license type is shown in Github. Currently it only says “License”.

PR #453, opened 2019-09-22
Fixed - Sub menu toggles doesn't work on mobile with one menu set to both locations

acosmin:
Fix for #365

Please test it before merging.

PR #468, opened 2019-09-22
UPDATE #461: Load index.js synchronously

### LINKED ISSUE #461: The modal and the search button doesn't work when not logged in
To reproduce :

1. Click on the search button and the menu button and make sure that they work.
2. Log out from WordPress, return to the front end; now the buttons should not work.

nielslange:
Fixes #461

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -211,7 +211,6 @@ if ( ! function_exists( 'twentytwenty_register_scripts' ) ) {
 		}
 
 		wp_enqueue_script( 'twentytwenty-js', get_template_directory_uri() . '/assets/js/index.js', array(), $theme_version, false );
-		wp_script_add_data( 'twentytwenty-js', 'async', true );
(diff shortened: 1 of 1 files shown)


acosmin:
You should also probably set to `true` `wp_enqueue_script()`'s `$in_footer`

nielslange:
@acosmin I thought that loading the `index.js` in the footer would cause this issue. That's why I haven't used `$in_footer = true` but kept `$in_footer = false`.

Ismail-elkorchi:
@nielslange  The change made in https://github.com/WordPress/twentytwenty/pull/453/commits/aa391fe21a4bb987bf14cc0b50cd58576bfbc9d2 fixes https://github.com/WordPress/twentytwenty/issues/461.

ianbelanger79:
Fixed in #453

Issue #276, opened 2019-09-14
Cover image overlay opacity should use a range control


celloexpressions:
The cover image overlay opacity is currently a select control with 11 options at increments of 10%. This would be more appropriate as a range (slider) control, because most users will not care about the numerical value of this option, and the range approach allows for a more-visual interaction. This should also be instant live-previewed in the customizer with `postMessage` for instant feedback.

![image](https://user-images.githubusercontent.com/7875961/64913376-dd514a80-d6f3-11e9-8baa-ab986d13ae33.png)

I'm not set up to create a PR yet, but this is the revised customize control code (someone else can feel free to make this into a PR):

```
/* Overlay Color Opacity --------- */
$wp_customize->add_setting(
	'twentytwenty_cover_template_overlay_opacity',
	array(
		'default'           => 80,
		'sanitize_callback' => 'absint',
	)
);

$wp_customize->add_control(
	'twentytwenty_cover_template_overlay_opacity',
	array(
		'label'       => __( 'Image Overlay Opacity', 'twentytwenty' ),
		'description' => __( 'Make sure that the value is high enough that the text is readable.', 'twentytwenty' ),
		'section'     => 'twentytwenty_cover_template_options',
		'type'        => 'range',
		'input_attrs' => array(
			'min'  => 0,
			'max'  => 100,
			'step' => 10,
		),
	)
)
```

The step of `10` matches the current options; I would suggest using a step of 5 instead for more-granular control. This also transitions the setting to be saved as an integer, and the output should be adjusted accordingly. Also note that the `settings` parameter in the control is not required when the control ID matches the setting ID.

joyously:
I think this is a good idea. Should the min be more than 0? Or max less than 100?

acosmin:
I could probably add this to #258 and also make the necessary changes for `parameter in the control is not required when the control ID matches the setting ID` where it's needed

carolinan:
I think this or some version of it will be part of https://github.com/WordPress/twentytwenty/issues/284 /#249

LittleBigThing:
> I think this is a good idea. Should the min be more than 0? Or max less than 100?

See #470 about this. I would go for 0 as a minimum, but limit it to less than 100%.

Issue #470, opened 2019-09-22
Consider limiting opacity for cover template to under 100%


LittleBigThing:
When the Cover Template page/post template is chosen, the featured image is overlayed with a customizable color layer. The color and the opacity of the overlay can be adjusted in the Customizer in a separate Cover Template section (currently).

The color overlay uses the [`mix-blend-mode` CSS property](https://caniuse.com/#feat=mdn-css_properties_mix-blend-mode) to create interesting filter effects for the image. This property is, however, not supported by IE and Edge. This means that the result is quite different in IE and Edge than expected. For example, adjusting the value for the property (for example, from multiply to hue or anything else) will not result in a different effect. This might be mentioned in the description to avoid confusion. Or we could drop the possibility to modify the value for the property.

Moreover, there can be a problem with the visibility of the featured image when the opacity is set to 100%. In this case, the image is completely covered by the color overlay in IE and Edge. The overlay opacity is set to 80% by default (`opacity: 0.8`), so this is less of a problem. To avoid completely covering the image, we could consider limiting the opacity setting to max. 90%.

LittleBigThing:
> Some browsers, like Internet Explorer and Edge, only support the "Normal" mode.

I just noticed that there is already a mention about non-supporting browsers when choosing the blend mode value in the Customizer. This is great and it could be the way to go. Although I did not notice it obviously. :-)

We might consider making 'normal' the default value instead of 'multiply'. This would give us defaults that are supported by all major browsers.

pattonwebz:
It could be cool to do browser support detection. on the frontend and adjust accordingly. We did not add any feature detection scripting to the theme however maybe we could handle it purely via CSS?

LittleBigThing:
Yes, we did something like that for Twenty Nineteen using [`@supports`](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports) in CSS. I am not sure it would help in this case since the user has a lot of freedom choosing the settings.

For Twenty Nineteen, there was a fixed setting with no opacity of the overlays, which resulted in an invisible featured image in IE and Edge. We applied some opacity for browsers not supporting `mix-blend-mode`. You can see it here: https://github.com/WordPress/twentynineteen/blob/b604f127c2cae10bd48bbbec0fbbbff2cd31f957/style.css#L2372
vs.:
https://github.com/WordPress/twentynineteen/blob/b604f127c2cae10bd48bbbec0fbbbff2cd31f957/style.css#L2377
[...]

PR #473, opened 2019-09-22
Range control for cover template opacity

LittleBigThing:
A range control to adjust the opacity of the cover template (step = 10 with min. 0 and max. 100)

fixing #276


joyously:
Why use a step of 10?
And there is concern about using max of 100...

LittleBigThing:
I agree to use steps of `5`. I just did a quick implementation based on your suggestion to make it testable. It might all still depend on decision about the opacity limits (#470)?

ianbelanger79:
I updated #470 with the latest information on this issue. You can resume work on this PR, with one suggestion, could we show the actual number you are selecting when using the range slider? Thanks

LittleBigThing:
Great, I will update the PR asap.

joyously:
I don't think the number should be shown. That's the whole point of using a range control.

acosmin:
@ianbelanger79 why was the blend mode option removed in this PR?

LittleBigThing:
@acosmin see this reply https://github.com/WordPress/twentytwenty/issues/470#issuecomment-534591170

We should have done it in 2 separate PRs maybe, if that is what you meant. I thought it was related enough.

acosmin:
@LittleBigThing thanks, I didn't know that.

ianbelanger79:
@LittleBigThing sorry, I should've seen this before merging. We also plan to remove the `mix-blend-mode` because it is not supported in IE11 or Edge

LittleBigThing:
Ok. So it’s just an overlay with opacity?
A separate PR will do it.

Issue #482, opened 2019-09-23
Replace array short syntax `[]` with `array()`


pattonwebz:
https://github.com/WordPress/twentytwenty/blob/bbfc3702bc621efe71e4b9a22e9911e99ba81c99/functions.php#L136

PR #484, opened 2019-09-23
#482 Update - Replace array short syntax `[]` with `array()`

tm-const:
Replaced short syntax with array


functions.php
```diff
@@ -133,7 +133,7 @@ function twentytwenty_theme_support() {
 		 * by the theme.
 		 */
 		$loader = new TwentyTwenty_Script_Loader();
-		add_filter( 'script_loader_tag', [ $loader, 'filter_script_loader_tag' ], 10, 2 );
+		add_filter( 'script_loader_tag', array( $loader => 'filter_script_loader_tag' ), 10, 2 );
```
pattonwebz: This should still remain in the format of `array($obj, 'myCallbackMethod')`.

```suggestion
		add_filter( 'script_loader_tag', array( $loader, 'filter_script_loader_tag' ), 10, 2 );
```

tm-const: Updated: https://github.com/WordPress/twentytwenty/pull/484/commits/72a427983ab8ae36d988d6b59898b6423a9550e8

Issue #483, opened 2019-09-23
JSHint errors need to be fixed


ianbelanger79:
Errors referenced here https://travis-ci.com/WordPress/wordpress-develop/jobs/238313146 need to be fixed

PR #485, opened 2019-09-24
Fixed JSHint errors

byalextran:
Fixes https://github.com/WordPress/twentytwenty/issues/483

One of the errors was about the function `twentyTwentyColor` being defined but never used. I deleted the function, however, it would be great if someone more familiar with the codebase could confirm that's the proper fix.


byalextran:
well, that's interesting. i removed the commas that were causing the errors. but the build fails now saying i need those commas. perhaps i'm misunderstanding what the fix involves?

acosmin:
I think it's because https://github.com/WordPress/wordpress-develop/blob/master/.jshintrc has `trailing` set to true and the default is false in this repo's eslint config...

ntwb:
That's indeed the case @acosmin, the ESLint config we use here is what is used in the Gutenberg repo and what will be used in WordPress core in the future (when Gutenberg comes time to be merged into core)

The best workaround here I see is to add the ESLint `comma-dangle` rule to this repo's ESLint config, I've added a commit that should resolve that.

ntwb:
There's another ~12 instances of JS files not previously checked that also require this, and a new end of file new line issue requires resolving, see https://travis-ci.com/WordPress/twentytwenty/jobs/238426502#L545

ianbelanger79:
> One of the errors was about the function twentyTwentyColor being defined but never used. I deleted the function, however, it would be great if someone more familiar with the codebase could confirm that's the proper fix.

Thanks for the PR @byalextran, it appears that the `twentyTwentyColor()` function is used in https://github.com/WordPress/twentytwenty/blob/932b16248efd6d142e928fa157fe187d7063859f/assets/js/customize-controls.js#L59

Also the core tests passed without removing the `twentyTwentyColor()` function from `assest/js/color-calculations.js` see https://core.trac.wordpress.org/changeset/46277

Could you please update your PR? Thanks

byalextran:
@ianbelanger79 @ntwb thanks for the feedback/insight. i'll update this PR later this evening.

ntwb:
@byalextran Thanks for the update, sadly I made a mistake in my ESLint rule change, it should be `never` not `always`.

The updated Travis build should report the correct errors now ;)

PR #506, opened 2019-09-24
Replace JS-based smooth scrolling with CSS.

### LINKED ISSUE #476: Consider removing JS-based smooth scroll
Smooth scroll is currently implemented using a JS approach, that hijacks all anchors pointing to another location in the page, and replaces their default behaviour with a script that animates a page scroll. This is problematic for a number of reasons:

- The animation is JS-based, blocking the main thread while executing and potentially leading to jank in slower devices (and some faster ones where JS-driven animations run at 60Hz while CSS-driven ones run at higher rates).
- Being JS-based, the behaviour only takes place after the JS is fully loaded and executed, leading to inconsistencies in behaviour between the initial interactive page and the final one.
- It hijacks regular anchors, which could interfere with screen readers and other assistive technologies.
- It's a non-configurable behaviour that may be undesirable for certain categories of users that prefer less motion.
- It's a non-standard behaviour that may interfere with how certain user agents display scrolling pages.

The current implementation is ~130 lines of complex JS (between smooth scroll, scrollToElement, and utility methods), that tries to take into account a number of things in the page to calculate the correct animation.

I propose instead using the [standard `scroll-behavior` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) to obtain a similar effect. Switching to `scroll-behavior` would eliminate all the above issues, while reducing the implementation to a single line of CSS:

```css
html {
  scroll-behavior: smooth;
}
```

This could be disabled for the purpose of other animations where the effect is not desirable (e.g. showing and closing modals) by adding/removing a single CSS class to the body:

```css
.disable-smooth-scrolling {
  scroll-behavior: auto;
}
```

The downside of this approach is that it's not supported by all browsers, but given that this is a small visual improvement, I think it would fall squarely in the category of progressive enhancement, for which a standard behaviour fallback is acceptable, and preferable to the JS-based approach.

sgomes:
Hello there! 👋 This PR fixes #476.

For full details, please read #476, but here's a short summary: the current JS-based approach has performance and accessibility flaws, and the proposed CSS-based approach here fixes most of the issues. It doesn't work in all browsers, but I propose that it doesn't have to, and can be viewed as a progressive enhancement. It delegates the implementation to the browser, which should make it as fast as native scrolling, and it allows for preferences like reduced motion to be honoured.

There appears to be no need for a modal workaround, unlike previous versions of the theme I tried this on, but please test thoroughly and let me know if there are any issues!

Some browsers with support for the approach in this PR:
- Recent versions of Chrome
- Recent versions of Firefox

Some browsers without support for the approach in this PR (revert to instant scroll):
- IE
- Safari


ntwb:
Unfortunately it doesn't look like there is enough browser support for `scroll-behavior` yet, see https://caniuse.com/#feat=mdn-css_properties_scroll-behavior

sgomes:
> Unfortunately it doesn't look like there is enough browser support for `scroll-behavior` yet, see https://caniuse.com/#feat=mdn-css_properties_scroll-behavior

Right, I do mention that in both the PR and the previous issue. Part of the argument is that I believe that this is not a critical feature, and can be seen as a progressive enhancement, something that will work better in the browsers that support it.

I don't believe having smooth scroll across all browsers justifies the issues with the current JS implementation.

sgomes:
Rebased to pick up some linting changes.

LittleBigThing:
Hi,

We may consider also adding support for people that opt out of having too much animation on their device for accessibility reasons:

```
@media screen and (prefers-reduced-motion: reduce) {
	html {
		scroll-behavior: auto;
	}
}
```

This disables smooth scroll according to the users preference.

ntwb:
@LittleBigThing, I've also been looking at that myself, possibly by adding the `stylelint-a11y` stylelint rule [media-prefers-reduced-motion](https://github.com/YozhikM/stylelint-a11y/blob/master/src/rules/media-prefers-reduced-motion/README.md)

sgomes:
> Hi,
> 
> We may consider also adding support for people that opt out of having too much animation on their device for accessibility reasons:
> 
> ```
> @media screen and (prefers-reduced-motion: reduce) {
> 	html {
> 		scroll-behavior: auto;
> 	}
> }
> ```
> 
> This disables smooth scroll according to the users preference.

Thanks, @LittleBigThing! I assumed the browsers did this themselves, since they have full control over the scroll animation, but that doesn't appear to be the case (at least with Chrome).

> @LittleBigThing, I've also been looking at that myself, possibly by adding the `stylelint-a11y` stylelint rule [media-prefers-reduced-motion](https://github.com/YozhikM/stylelint-a11y/blob/master/src/rules/media-prefers-reduced-motion/README.md)

I would recommend against that specific rule. `transition: none` can break certain animation sequences if they're relying on the `transitionend` event to be dispatched, which it won't be if there's no transition. One example of this is the recently added main menu animation.

I would instead opt for `transition: 1ms`, or any sufficiently low value that's perceived as instant, but still fires all the appropriate events.

sgomes:
Rebased on master.

LittleBigThing:
I think that style-rtl.css should also changed in de PR.

sgomes:
> I think that style-rtl.css should also changed in de PR.

You're right, I totally missed that file! Thanks!

ntwb:
See also @afercia's commment on Gutenberg here https://github.com/WordPress/gutenberg/pull/17645#issuecomment-536502715

ntwb:
The `scroll-behavior: auto;` change is helpful for Chrome & Firefox users, but Safari doesn't support `scroll-behavior`, there's wider browser support for `transition-duration`, though `prefers-reduced-motion` is also not supported on IE/Edge.

I'm honestly not sure what direction to take here... There's pros and cons on each method

https://caniuse.com/#feat=mdn-css_properties_scroll-behavior
https://caniuse.com/#feat=mdn-css_properties_transition-duration
https://caniuse.com/#feat=mdn-css_at-rules_media_prefers-reduced-motion

sgomes:
> The `scroll-behavior: auto;` change is helpful for Chrome & Firefox users, but Safari doesn't support `scroll-behavior`, there's wider browser support for `transition-duration`, though `prefers-reduced-motion` is also not supported on IE/Edge.
> 
> I'm honestly not sure what direction to take here... There's pros and cons on each method
> 
> https://caniuse.com/#feat=mdn-css_properties_scroll-behavior
> https://caniuse.com/#feat=mdn-css_properties_transition-duration
> https://caniuse.com/#feat=mdn-css_at-rules_media_prefers-reduced-motion

I see these as different things to be considered. Let me frame them in the form of individual questions that need decisions:

**1. Do we think the problems with the JS version are severe enough to invalidate it?**

I would argue so, but that's just my personal opinion.

**2. Do we want every browser to support smooth scroll?**

If we do, then we have to keep the JS-based approach. We could potentially implement both and use the CSS-based one if the user's browser supports it, falling back to JS if not.

**3. Do we want it to look and feel the same in every supported browser?**

If we do, then the JS-based approach is the only option, since browsers are free to implement the CSS-based option with some degree of freedom.

**4. Do we want to default to motion or non-motion, in browsers without support for the `prefers-reduced-motion`? feature**

Both are possible with either the CSS-based approach, or the JS-based approach.

LittleBigThing:
I agree with the current implementation of the PR. I think it is a great example of a progressive enhancement.

pattonwebz:
I moved this ticket over to trac here: https://core.trac.wordpress.org/ticket/48551#ticket

PR #510, opened 2019-09-24
UPDATE #509: Replace "Twenty Nineteen" with "Twenty Twenty"

### LINKED ISSUE #509: Replace "Twenty Nineteen" with "Twenty Twenty" in various comments
- [x] functions.php:117
- [x] functions.php:118
- [x] entry-header.php:6
- [x] featured-image.php:6

nielslange:
Fixes #509

### DIFF
--- a/classes/class-twentytwenty-svg-icons.php
+++ b/classes/class-twentytwenty-svg-icons.php
@@ -10,7 +10,7 @@
 if ( ! class_exists( 'TwentyTwenty_SVG_Icons' ) ) {
 	/**
 	 * SVG ICONS CLASS
-	 * Retrieve the SVG code for the specified icon. Based on a solution in TwentyNineteen.
+	 * Retrieve the SVG code for the specified icon. Based on a solution in Twenty Nineteen.
 	 */
 	class TwentyTwenty_SVG_Icons {
 		/**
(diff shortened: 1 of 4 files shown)


classes/class-twentytwenty-svg-icons.php
```diff
@@ -10,7 +10,7 @@
 if ( ! class_exists( 'TwentyTwenty_SVG_Icons' ) ) {
 	/**
 	 * SVG ICONS CLASS
-	 * Retrieve the SVG code for the specified icon. Based on a solution in TwentyNineteen.
+	 * Retrieve the SVG code for the specified icon. Based on a solution in Twenty Nineteen.
```
ianbelanger79: This one should probably say Twenty Twenty

nielslange: I thought so too initially, but it says _"Based on a solution in Twenty Nineteen."_ Thus, it refers to the source of that code snippet. There's one more similar comment in `classes/class-twentytwenty-walker-comment.php:13`:

> A custom walker for comments, based on the walker in Twenty Nineteen.

As both parts referring to the source, I would not change them.

ianbelanger79: I should've read the whole thing ;) thanks for the clarification.

Issue #511, opened 2019-09-24
The cover on pages gets cut-off


acosmin:
The cover (cover page template) on pages gets cut-off at the bottom when the admin bar is present.

ianbelanger79:
Duplicate of #489

PR #512, opened 2019-09-24
Fix - The cover on pages gets cut-off at the bottom

acosmin:
When the admin bar is present.

Fix for #511


acosmin:
I'll try and remake the PR :) I think I messed something up...

acosmin:
Remade in #520

Issue #522, opened 2019-09-25
Global JS variables should be prefixed


acosmin:
https://github.com/WordPress/twentytwenty/blob/48edb092fec7710ef5436dd84f39b5c02081c570/functions.php#L581-L582

PR #523, opened 2019-09-25
Fix - Global JS variables should be prefixed

acosmin:
Prefixed  `previewElements` and `backgroundColors`

Fix for #522


ianbelanger79:
Closed due to merge conflicts, fixed in redo #536

Issue #524, opened 2019-09-25
Author Bio - Customizer option to show or hide


acosmin:
As the title says, we should add an option in the Customizer to enable or disable this single view section.

karmatosed:
This could be personal, but I really would encourage less options not more. With regards to turning off / on an author bio, this falls into an extra not needed option.

acosmin:
An extra is also the `Author Bio` section, it's not a core thing and I think having a way to disable it is needed. The only other way you would be able to disable it is to - edit user > remove any `Biographical Info` (this also applies to one user, not for all of them)

jarretc:
Casting my vote for this issue. As somebody that frequents the themes support forum, the more options we give users that allow them to customize their site the better. Extra options don't hurt anybody, plus it gives new theme developers plenty of data to utilize for making their own themes by looking at how the code is implemented in the default themes.

PR #525, opened 2019-09-25
Fix - Author Bio - Customizer option to show or hide

acosmin:
Fix for #524


LittleBigThing:
I think it is great to have this option since not everyone wants to show this information.

IMHO:

- We should change the description from

> 'Uncheck to hide the author bio section in posts.'

to

> 'Check to show the author bio section in posts.'

- possibly add an `active_callback` for single posts to only show the setting on single posts, and make the change observable.

pattonwebz:
At this point we won't be adding any more options untill after release but I am tagging this one as `future` to be considered after the release.

The code here works fine, I tested when PR was created. A decision is needed if we should add more options (perhaps if we don't want any options added then a filter override could be added instead).

acosmin:
A filter can work too 😄 I can change it to that if it's wanted...

PR #527, opened 2019-09-25
Fix left/right alignment of Image block

### LINKED ISSUE #526: Image blocks with left/right alignment are not positioned correctly
When a Image block is aligned to the left or right, they float to the left or right within the entry-content column instead of being positioned beside it, like the other alignleft/alignright elements.

This is caused by the `block-styles` CSS resetting the horizontal margins of `.wp-block-image .alignright` and `.wp-block-image .alignleft` to 0. The fix is to update the `.alignleft` and `.alignright` styles in the media queries targeting the `660px` and `1280px` breakpoints to target `.entry-content .alignleft` and `.entry-content .alignright, taking priority over the CSS in  `block-styles`.

**Current behavior:**
image

**Expected behavior:**
image

andersnoren:
Fixes #526.

### DIFF
--- a/style.css
+++ b/style.css
@@ -4302,11 +4302,11 @@ a.to-the-top > * {
 
 	/* ALIGNMENT CLASSES */
 
-	.alignleft {
+	.entry-content .alignleft {
 		margin-left: calc(( 100vw - 58rem - 8rem ) / -2);
 	}
 
-	.alignright {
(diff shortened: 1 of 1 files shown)


joyously:
Does this problem affect the editor at all? (or it can't get wide enough to be affected?)

PR #534, opened 2019-09-25
Fix uneven post spacing on archive

### LINKED ISSUE #531: Uneven spacing between posts on archive pages
Posts on archive pages have both a top-padding and a separator with a margin, resulting in the top spacing being twice as big as the bottom spacing for all but the first post on each page.

image

This can be fixed by changing the `body:not(.single) .post` targeting to `body:not(.single) .post:first-child` in the two places where it is used.

andersnoren:
Updates the targeting to only hit the first article element, since the separator elements provide spacing between posts. Also updates the targeting to hit article elements that display pages or other non-post CPTs (in search results or on archive pages for custom post types).

Fixes #531.

### DIFF
--- a/style.css
+++ b/style.css
@@ -2400,7 +2400,7 @@ body.template-cover .entry-header {
 	margin: 0;
 }
 
-body:not(.single) .post {
+body:not(.singular) main article:first-of-type {
 	padding: 4rem 0 0;
 }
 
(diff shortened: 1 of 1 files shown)


joyously:
Are you aware that there is a core body class for `archive` and `search`? That would match better for `:not(.singular)`.
You can also put a generic `entry` class on the article element and target all post types easily.

Issue #537, opened 2019-09-25
Use selective refresh for cover overlay opacity and fixed background image


acosmin:
These items should use selective refresh.

PR #538, opened 2019-09-25
Fix - Use selective refresh for cover overlay opacity and fixed background image

acosmin:
Fix for #537

Best viewed when a featured image is used.

### DIFF
--- a/classes/class-twentytwenty-customize.php
+++ b/classes/class-twentytwenty-customize.php
@@ -353,6 +353,7 @@ if ( ! class_exists( 'TwentyTwenty_Customize' ) ) {
 				array(
 					'default'           => 'multiply',
 					'sanitize_callback' => array( __CLASS__, 'sanitize_select' ),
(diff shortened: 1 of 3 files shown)


acosmin:
I guess I'll have to fix some conflicts 😄

classes/class-twentytwenty-customize.php
```diff
@@ -464,3 +472,28 @@ function twentytwenty_customize_partial_site_logo() {
...
+	 *     The attributes
+	 * 
+	 *     @type int $min Minimum value
+	 *     @type int $max Maximum value
+	 *     @type int $step Interval between numbers
+	 * }
+	 */
+	return apply_filters( 'twentytwenty_customize_opacity_range', array(
+		'min'  => 0,
+		'max'  => 100,
```
ianbelanger79: Max should only be 90

acosmin: done

PR #544, opened 2019-09-25
UPDATE #370: Keep inputs LTR when site is RTL

### LINKED ISSUE #370: RTL: some inputs should be LTR
I am hoping someone knowledgeable about RTL can confirm or elaborate, but I learned recently that some input fields should always be LTR. This is for telephone numbers, email, and URLs.
Something like
```
.rtl input[type="url"],
.rtl input[type="email"],
.rtl input[type="tel"] {
	direction: ltr;
}
```

nielslange:
Fixes #370


style-rtl.css
```diff
@@ -813,6 +813,13 @@ textarea {
 	border-color: #dcd7ca;
 }
 
+code,
+input[type="url"],
+input[type="email"],
+input[type="tel"] { /* stylelint-disable-line */
+	direction: rtl;
```
carolinan: Is this supposed to say ltr?

ianbelanger79: It's in the rtl stylesheet, so I assume that it's supposed to be rtl

nielslange: Yes, it should say `LTR` instead of `RTL`. I'll double-check how this could happen, @carolinan. Thanks for pointing it out! 🙌

nielslange: It seemed the combinations of build and lint statements fails here:

```
code,
input[type="url"],
input[type="email"],
input[type="tel"] {
	/*rtl:ignore*/ /* stylelint-disable-line */
	direction: ltr;
}
```

I fixed it by changing the part to:

```
code,
input[type="url"],
input[type="email"],
input[type="tel"] {

	/*rtl:ignore*/
	direction: ltr;
}
```

nielslange: @ntwb I'm sure you can assist me to solve the following mystery. First I tried

`/*rtl:ignore*/ /* stylelint-disable-line */`

which failed. Then I tried 

`/* rtl:ignore stylelint-disable-line */`

which also failed. Last but not least, I tried

```
/*rtl:ignore*/
/* stylelint-disable-line */
```

which, again, failed. I now ended up with the following code:

```
code,
input[type="url"],
input[type="email"],
input[type="tel"] {

	/*rtl:ignore*/
	direction: ltr;
}
```

That code is far from elegant and while it works, it does not look correct to me. Can you tell me how to avoid the empty line while still respect linting?

Issue #185, opened 2019-09-10
What is the use of the Full Width Template?


yannickiki:
I don't see any difference between Default Template and Full Width Template. Is it a leftover from Chaplin?

andersnoren:
@yannickiki Thanks for posting the issue! I think there's an expectation for a full width template still (even if the block editor now allows for a lot of the modifications they've often been used for), so we probably want to include one. It seems like no styles are being applied to it currently though, like you say.

The full-width template should:

* Remove the `.thin` class from the wrapping `.post-inner` element, so the content container extends to the width of the layout.
* Remove the max-width values from `.entry-content` and `.entry-content p`, so they extend to the width of the container.
* When we implement the `.alignleft` and `.alignright` classes with negative horizontal margins, positioning them outside the main content column (the pull quote in [this image](https://make.wordpress.org/core/files/2019/09/twenty-twenty-single-post.jpg)), we'll need to adjust those classes to make sure they don't extend outside the main content column on the full width template.

LittleBigThing:
I have asked a similar question here: https://github.com/WordPress/twentytwenty/issues/497#issuecomment-534942966

Also, do these styles still have to be applied?
Could we make the differences or the benefits using the template clear in some way?

byalextran:
does https://github.com/WordPress/twentytwenty/pull/527 address `.alignleft` and `.alignright` as you described above?. 

if so, and assuming my PR is valid, i think additional styling tweaks are needed.

![Screen Shot 2019-09-25 at 22 23 03 PM](https://user-images.githubusercontent.com/455406/65655889-70875b80-dfe3-11e9-9abe-feedb947b2a7.png)

![Screen Shot 2019-09-25 at 22 23 13 PM](https://user-images.githubusercontent.com/455406/65655925-8bf26680-dfe3-11e9-922b-c36a78745aec.png)

PR #547, opened 2019-09-26
Made content full width when using Full Width template

byalextran:
Fixes https://github.com/WordPress/twentytwenty/issues/185

Took at a stab at it. I assume any changes in `style.css` should be reflected in `style-rtl.css` as well?

### DIFF
--- a/style-rtl.css
+++ b/style-rtl.css
@@ -2356,6 +2356,10 @@ body.template-cover .entry-header {
 /*	7c. Template: Full Width
 /* -------------------------------------------------------------------------- */
 
+body.template-full-width .entry-content,
+body.template-full-width .entry-content p {
+	max-width: none;
+}
 
 /* -------------------------------------------------------------------------- */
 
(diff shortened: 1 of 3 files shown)


Aetles:
I have a quick question about this change: 

`<?php echo is_page_template( 'templates/template-full-width.php' ) ? '' : 'thin'; ?>`

This means that there can only be one full width template, since any other template [that still loads `content.php`] will have the `thin` class added, right?

So if I create a child theme for Twenty Twenty (as many others will now that this is the new default theme) and I want to add my own full width template I have to deal with the added `thin` class in some way. Either by redefining the class with CSS or add the `content.php` template to my child theme as well and change it. (Not that it is _that_ much work, but I guess some will find it confusing maybe before they figure it out.)

joyously:
Yes, that is unfortunate. We tried to factor out the page templates to be separate, but I guess the solution that remains didn't fulfill that.
My recommendation would be to call a filter for the class in the page template instead of checking for the page template by name. And make the default have no class, and the exception have a class (so it would be the reverse of what is there).

byalextran:
would addressing it be considered a bug fix and allowed in the RC stage or would it be considered an enhancement that has to wait until after 5.3 releases?

PR #548, opened 2019-09-26
UPDATE #442: Fix missing :hover style

### LINKED ISSUE #442: Some links are missing a :hover style
Links need to have a visual change when hovered.

Most of the links that are missing a hover style are in the content,
where the link is always underlined:
Links in post and page content
Links in captions, both in blocks and normal.
Blocks examples:
Links in cover block text (both image and video)
Links in quote blocks
Links in media & text block

_Edge case:_
Linked images in post content is missing any indication except the mouse pointer change.

----
But then we have some links in blocks that are not underlined in their default state, 
but underlined when hovered. The opposite of the other links in the content.

Archive and category widgets
Latest comments widget

Latest posts widget:
This one is different, the first item of the latest posts when in list view has _no underline,_
but the rest of the items have an underline when hovered.

----

Page breaks
Edit this
Category links

Comment author name
Links in comment content

Horizontal menu: 
current-menu-item is always underlined.

nielslange:
Fixes #442


nielslange:
@andersnoren No worries. I'll solve the merge conflict and adjust the other two issues.

style-rtl.css
```diff
@@ -2463,6 +2476,12 @@ body:not(.singular) main article:first-of-type {
 	text-transform: uppercase;
 }
 
+.entry-categories a:focus,
+.entry-categories a:hover {
+	text-decoration: underline;
+	border-bottom: none;
```
andersnoren: You can remove the `text-decoration: underline` here, and set it to `border-bottom-color: transparent` instead, so the elements beneath the link doesn't jump when the border width gets removed.

nielslange: @andersnoren With https://github.com/WordPress/twentytwenty/pull/548/commits/5d74df15fea5118d386a41e65eb7ab756b58f4de I've implemented your feedback. Feel free to test this PR again. 😀

PR #557, opened 2019-09-26
Fix animated menus.

### LINKED ISSUE #513: Sub-menus are not animating
Looking at the source JS, there appears to be an intention to animate sub-menu opening and closing with a slide animation. However, they appear to be opening and closing instantly, with no animation (animations at 25% speed in the GIF below).

ezgif-4-224772654588

This should be fixed by either:
1 - Deciding not to animate, and removing the animation code from the JS sources.
2 - Fixing the animation.

I'm happy to implement whichever fix is deemed the best.

sgomes:
The previous implementation didn't work, and even if it did, it would have performance issues due to animating layout-invalidating properties like `height`.

This rewrite animates the menu using only opacity and transform, leading to smooth animations in most devices. The animation is JS-orchestrated, but entirely CSS-driven. The animation is skipped if a user has configured their OS or browser to reduce animations.

Note that the implementation is complex, but this is required complexity for smooth, cross-browser animations in a sliding menu scenario. When reviewing, **please consider whether this complexity is desirable, or whether it would be preferable to drop animations instead**.

This change also fixes issues with automatically opening sub-menus on page load.

Summary of changes:
- Rewrite menu animation functionality.
- Extract toggling function out of `click` event hander, so it can be used without triggering click events.
- Ensure that the initial menu opening is done on all menus, not just the first one (there are two menus; one for desktop and one for mobile).

Fixes #513. Tested and working in recent Chrome, recent Firefox, recent Safari, and IE11.

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -357,17 +357,20 @@ twentytwenty.modalMenu = {
 	},
 
 	expandLevel: function() {
-		var modalMenu = document.querySelector( '.modal-menu' );
-		var activeMenuItem = modalMenu.querySelector( '.current-menu-item' );
+		var modalMenus = document.querySelectorAll( '.modal-menu' );
 
-		if ( activeMenuItem ) {
-			twentytwentyFindParents( activeMenuItem, 'li' ).forEach( function( element ) {
-				var subMenuToggle = element.querySelector( '.sub-menu-toggle' );
-				if ( subMenuToggle ) {
-					subMenuToggle.click();
-				}
-			} );
-		}
+		modalMenus.forEach( function( modalMenu ) {
+			var activeMenuItem = modalMenu.querySelector( '.current-menu-item' );
+
+			if ( activeMenuItem ) {
+				twentytwentyFindParents( activeMenuItem, 'li' ).forEach( function( element ) {
+					var subMenuToggle = element.querySelector( '.sub-menu-toggle' );
+					if ( subMenuToggle ) {
+						twentytwenty.toggles.performToggle( subMenuToggle, true );
+					}
+				} );
+			}
+		} );
 	}
 }; // twentytwenty.modalMenu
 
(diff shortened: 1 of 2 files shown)


acosmin:
nice, works as described...

andersnoren:
@sgomes The animation itself looks and feels great – nice work! My only concern is with how the social menu behaves when the sub menus are expanded and collapsed, as in the gif below.

ezgif com-video-to-gif

Would it be possible to hide the `.menu-bottom` element beneath the `.menu-top` element when the sub menus are expanding and collapsing, so it doesn't "flash" in its original position when toggling?

sgomes:
> @sgomes The animation itself looks and feels great – nice work! My only concern is with how the social menu behaves when the sub menus are expanded and collapsed, as in the gif below.
> 
> ezgif com-video-to-gif
> 
> Would it be possible to hide the `.menu-bottom` element beneath the `.menu-top` element when the sub menus are expanding and collapsing, so it doesn't "flash" in its original position when toggling?

Excellent catch, @andersnoren! I'll see what I can do about that one 👍

sgomes:
@andersnoren: Pushed a fix that animates the menu bottom with the rest of the menu 🙂

This should work everywhere, except in IE, where the menu bottom already appears to have issues when the content overflows.

sgomes:
> That. Looks. Awesome. 👌
> 
> Adding `flex-shrink: 0;` to the `.menu-top` element (around line 2050 in `style.css`) should take care of the flexing issue in IE11 as well.

Woohoo, thank you! I'd been chasing this for hours with much more complicated ideas.

That does do the trick nicely, when combined with an IE-only overflow fix on the container 👍

andersnoren:
@sgomes What's your WordPress.org username, btw? For the contributors.md file.

sgomes:
Woohoo, thank you for all your help, @andersnoren! 

> @sgomes What's your WordPress.org username, btw? For the contributors.md file.

It's `sergiomdgomes`. Thanks!

PR #561, opened 2019-09-26
Add `:focus` pseudo-class selector to `:hover` selectors

ntwb:
Updates `:hover` selectors with a `:focus` pseudo-class selector

This is part of a pull request to follow adding the `a11y/selector-pseudo-class-focus` [stylelint-a11y rule](https://github.com/YozhikM/stylelint-a11y/tree/master/src/rules/selector-pseudo-class-focus) to check for accessibility issues

See https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

### DIFF
--- a/style.css
+++ b/style.css
@@ -903,22 +903,22 @@ input[type="submit"] {
 	transition: opacity 0.15s linear;
 }
 
-button:hover,
-.button:hover,
-.faux-button:hover,
-.wp-block-button__link:hover,
-.wp-block-file .wp-block-file__button:hover,
-input[type="button"]:hover,
-input[type="reset"]:hover,
-input[type="submit"]:hover,
 button:focus,
+button:hover,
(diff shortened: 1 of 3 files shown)


style.css
```diff
@@ -1109,20 +1109,23 @@ button.toggle {
 /* COLOR */
 
 .color-accent,
+.color-accent-hover:focus,
```
andersnoren: You need to update the targeting for ['content']['accent'] in`twentytwenty_get_elements_array()` to also include the new helper class targets: `.color-accent-hover:focus`, `.bg-accent-hover:focus` and `.border-color-accent-hover:focus`.

ntwb: Thanks, done in c2c64697def9fd8cc0c42f2bc34d408d99ceaabd

PR #577, opened 2019-09-27
UPDATE #573: Remove button style "Squared" from button block

### LINKED ISSUE #573: Remove button style "Squared" from button block
The button block comes with the following three styles:

- Default → Rounded button
- Outline → Rounded, outlined button
- Squared → Squared button

With Twenty Twenty we show only squared buttons. Therefore, the `Default` and the `Squared` button styles look identical, which might be confusing for some users. To avoid confusion I suggest removing the `Squared` button style.

nielslange:
Fixes #573

<table>
<tr>
<td>Before:
<br><br>

#573 - before

</td>
<td>After:
<br><br>

#573 - after

</td>
</tr>
</table>


assets/js/editor-script-block.js
```diff
@@ -1,8 +1,9 @@
-/** 
+/**
  * Remove squared button style
- * 
+ *
  * @since 1.0.0
-*/
-wp.domReady( () => {
+ */
+/* eslint-disable no-undef */
```
nielslange: @ntwb, @aristath or @fabiankaegy: As I've seen the warning `'wp' is not defined.eslint(no-undef)` I've added `/* eslint-disable no-undef */`. Is this the correct way of handling it? I do have to say that I'm not very familiar with es-lint, that's why I reaching out to you.

aristath: It's preferable if you just add this at the top of the file:
```
/* global wp, _ */
```
You can add others there too. In my example I'm defining wp and underscores, but you can add the ones you need comma-separated

nielslange: Great! Thanks for your reply, @aristath! Great to learn how to handle these issues properly on the go. 😀

nielslange:
@aristath I'm sure you can also answer another question I had earlier. Initially, I used

```
wp.domReady( () => {
	wp.blocks.unregisterBlockStyle( 'core/button', 'squared' );
} );
```

but that led to an error. Now, I'm using 

```
wp.domReady( function() {
	wp.blocks.unregisterBlockStyle( 'core/button', 'squared' );
} );
```

Are we using ES5 and are arrow functions not allowed yet? I couldn't find any information regarding this in https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/ though.

aristath:
Yeah, arrow functions are not supported in ES5.

assets/js/editor-script-block.js
nielslange: @aristath With https://github.com/WordPress/twentytwenty/pull/577/commits/ea36ed9c4a5542dcfb8c866007b7ed7bcc6566ab I brought this PR in proper shape. 😀

Issue #542, opened 2019-09-25
Our Fonts folder is too big, we need a good solution to cut it down


ianbelanger79:
We are using the Inter variable font, which is great for most browsers. However, since we have to support IE11 and earlier versions of Edge, we must include over 20 Inter font files.

We need to figure out a good solution to cut down our current themes folder which is over 3MB.

ianbelanger79:
IE11 without Inter font
![ie11-without-inter](https://user-images.githubusercontent.com/34601694/65632921-d73b5380-dfa8-11e9-9ed2-a97de1fd6f78.PNG)

IE11 with Inter font
![ie11-with-inter](https://user-images.githubusercontent.com/34601694/65632930-df938e80-dfa8-11e9-95b7-6e210633ae9a.PNG)

andersnoren:
We've discussed only including the variable fonts in the theme, and using a stack of system fonts in browsers that do not have support for variable fonts ([list here](https://caniuse.com/#search=variable%20fonts)). That would bring the size of the Inter folder down to 470 KB.

ntwb:
I presume #492 reduced the size of the font-stack somewhat?

nielslange:
@ntwb I'm afraid that https://github.com/WordPress/twentytwenty/issues/492 didn't really decrease the total font size.

These will be the folder sizes once #118 got merged:

```
160K    Cairo
400K    Noto_Sans
5.6M    Noto_Sans_HK
4.3M    Noto_Sans_JP
2.1M    Noto_Sans_KR
7.9M    Noto_Sans_SC
5.4M    Noto_Sans_TC
148K    font-awesome
3.0M    inter
```

Please note that the following fonts will only be loaded as a non-latin fallback font when required:

- Cairo
- Noto_Sans
- Noto_Sans_HK
- Noto_Sans_JP
- Noto_Sans_KR
- Noto_Sans_SC
- Noto_Sans_TC

cc: @ianbelanger79 & @andersnoren

joyously:
There was someone in #core today, saying that they tried to install the Beta to test, and it timed out halfway through the download (looked like ~10meg out of ~20meg). The zip for WordPress has increased in size with every release, but it shouldn't be larger mostly because of the theme's fonts!

Please just use the fonts that the admin already uses or the generic `sans-serif` and `serif` that work for every language (the user gets what they already have chosen in their browser).

nielslange:
[...]

PR #581, opened 2019-09-27
Remove non-variable versions of Inter

andersnoren:
After the discussions in #542, this PR removes all versions of Inter except the variable font version. This brings the file size of the Inter folder down from 3.2 MB to 471 KB. This isn't necessarily the way we choose to go in the end, but it will help us reduce the file size of the theme ahead of Beta 2.

With this PR, all browsers that support variable fonts will get the variable fonts version of Inter, and all browsers that lack support for variable fonts get a system stack of fonts. You can find a list of all browsers that support variable fonts support [here](https://caniuse.com/#search=variable%20fonts).


style.css
```diff
@@ -121,7 +121,7 @@ body {
 	-moz-box-sizing: border-box;
 	box-sizing: border-box;
 	color: #000;
-	font-family: Inter, -apple-system, BlinkMacSystemFont, "Helvetica Neue", Helvetica, sans-serif;
+	font-family: -apple-system, BlinkMacSystemFont, "Helvetica Neue", Helvetica, sans-serif;
```
ianbelanger79: Should we be replacing "Inter" with "Inter var", instead of just removing it?

andersnoren: We include Inter var within the `@supports` media query in the next CSS rule, which only targets browsers that supports variable fonts.

PR #589, opened 2019-09-27
UPDATE #588: Update print styles

### LINKED ISSUE #588: Update print styles
The link href need to be hidden, because when printed this way, it can create a text overlap:

print2

nielslange:
Fixes #588


carolinan:
@Nielsange what is happening with this PR? 🙂

Issue #583, opened 2019-09-27
No way to reset the accent color hue


melchoyce:
Our colorpicker library support resetting the default colors, so we get that by default, but there's no way to reset the accent color back to the default value. Might be worth considering adding a "default" button, if possible.

ianbelanger79:
@aristath is it possible to have a default reset button for the hue picker?

aristath:
Unfortunately the WP colorpicker doesn't have a reset button/functionality for hue-only pickers.
It's doable, but we'd have to custom-code it... Like for example by injecting a reset button via JS right next to the slider (or above/below it), and then on click reset the control to its initial value.

joyously:
@aristath This is where you whip out your Reset-All button!  (although the one you wrote was for child themes, it would be great to get that into circulation)

nielslange:
I was also looking for the `Reset` button earlier today.

karmatosed:
I would like to add that I just experienced this and wanted a reset. I wonder if there is any option of getting it?

aristath:
> I would like to add that I just experienced this and wanted a reset. I wonder if there is any option of getting it?

We _can_ add it, but it's not something that is normally supported by that control.
So we really only have 3 choices here:
1. Don't do it.
2. Do it in wp-core and add a reset button in hue-only colorpickers
3. Custom-code it in the theme

It's too late for option 2 in WP5.3, so we can either custom-code this in the theme, or just implement it in wp-core in a future release. But it we do custom-code it in the theme then we'll need a mockup.

Or we could just not do it 😄 

--------------

Alternative suggestion: Add a "custom accent color" checkbox. The hue colorpicker will then be hidden by default and only revealed if that checkbox is checked.
If the checkbox is disabled, then just fallback to default.

acosmin:
I made a small prototype in #613 based on aristath's alternative suggestion.

ianbelanger79:
This was fixed in #613

PR #613, opened 2019-09-28
Accent color hue reset on active callback

acosmin:
A prototype for the suggestion made in #583

The Accent color hue control will activate/deactivate based on a checkbox. Also, it will reset its value on deactivate.


aristath:
I like it 👍

nielslange:
@acosmin Functionality wise I love your prototype. However, from a usability point of view, wouldn't it be possible to have a simple reset button below the `Accent Color Hue` slider that sets the colour back to `#a73050`?

acosmin:
@nielslange it is possible 😄 @aristath described it best in the issue mentioned above.

aristath:
> However, from a usability point of view, wouldn't it be possible to have a simple reset button below the `Accent Color Hue slider` that sets the colour back to `#a73050`?

This is one of those weird issues where we wave to tread carefully...
Yes, UX-wise it would be better to just have a reset button.
But that would require 1 of 2 things:
1. Build a custom control to override the default WP Color control, and add the reset button there.
2. Inject it via JS

I don't know if adding a custom control would be recommended considering this is a core theme, and injecting via JS seems hacky and ugly (code-wise).

It's a thin line we have to walk.

The above implementation may not be the best choice for UX, but it's the most native choice and "the WP way". Unfortunately "_the WP way_" usually means doing things we're not happy with to get around core limitations...

nielslange:
@aristath Thanks for your detailed explanation. 😀

joyously:
Which way did Twenty Seventeen and Twenty Nineteen do it?

nielslange:
That was a great question, @joyously! This is how Twenty Nineteen handled the same colour issue:

Screen Capture on 2019-09-28 at 21-58-51

@aristath and @acosmin As Twenty Nineteen has that functionality it shouldn't be too hard to implement it in the same way, right? This would also be great for folks switching from Twenty Nineteen to Twenty Twenty as they can reuse already learned knowledge.

aristath:
Well... The fact that twentynineteen did it that way doesn't necessarily mean that it's better than a checkbox.
They are both viable options.

acosmin:
I can do it either way, I would prefer this option, but ...

The project leads should decide if something like this is wanted.

nielslange:
> Well... The fact that twentynineteen did it that way doesn't necessarily mean that it's better than a checkbox. They are both viable options.

I agree with you. To me personally, the Twenty Nineteen approach is less confusing. But let's see how @andersnoren thinks about this. 😀

carolinan:
As per the meeting, we want to use the radio buttons.

acosmin:
@carolinan as requested 😄

acosmin:
@andersnoren done 👍

acosmin:
@ianbelanger79 I think this one can be merged 😄

PR #623, opened 2019-09-28
Reduce spacing when the same background color is set

### LINKED ISSUE #501: Example of the front page of Twenty Twenty theme not using cover template
Here is an example of Twenty Twenty theme not using the cover template but the default page template.

Notice the big Front page title and how much space it takes up. The standard page content using a Gutenberg paragraph and an image float has very little space below.

Twenty-Twenty-Front

A new suggestion. After some CSS modifications are added

Screen Shot 2019-09-24 at 13 08 48

It makes the Front page title smaller. It takes up less space below the header. Gutenberg content is wider.

andersnoren:
Adds styles to adjust the spacing between some elements when the "Background Color" and "Header & Footer Background Color" settings are set to the same colors, using the `.reduced-spacing` class added in #582.

Combined with updates to the alignment classes in previous PRs, fixes #501.

**Before:**
image

**After:**
image


dianeco:
Hi @andersnoren, maybe the spacing on the archive page for "archive-header" need to be adjusted?

andersnoren:
Never a good facepalm gif when you need one. Thanks @dianeco, good catch 🙂

PR #628, opened 2019-09-28
Update the mobile menu to use the close button

### LINKED ISSUE #625: Consider adding a close button for mobile menu to improve user experience and usability
Currently on mobile, to open the menu the user has to click on the menu button. To close it, the user has to click on the same button. Nothing clearly indicates that clicking on the same button will close the menu.

modal-mobile

On desktop, the user experience is different, to close the menu there is a “Close Menu” label with a close icon. There is no ambiguity.

modal-desktop

I suggest for a better user’s understanding and experience to show the Close Nav Toogle button on mobile and open the modal on top of the header. 

For design clarity, I don’t think it’s necessary to show the header (logo + search icon) when the modal is opened. Keeping the branding isn’t needed while browsing the menu.

modal-mobile-proposition

Here is a comparison:

comparison

andersnoren:
This PR updates the mobile menu to cover the entire screen, and use the close button which is also displayed on desktop. Fixes #625.

**Before and after:**
mobile-menu-update

### DIFF
--- a/style.css
+++ b/style.css
@@ -1499,7 +1499,6 @@ ul.social-icons li {
 #site-header {
 	background: #fff;
 	position: relative;
-	z-index: 9999;
 }
 
 .header-inner {
@@ -1901,7 +1900,6 @@ ul.primary-menu {
 	display: none;
 	opacity: 0;
 	overflow: auto;
-	padding: 8.4rem 0 0 0;
 	position: fixed;
 	bottom: 0;
 	left: -99999rem;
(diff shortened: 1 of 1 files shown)


yannickiki:
Hi @andersnoren the code to display the close nav toogle is only executed where there is an expanded menu. The condition should be removed in the modal-menu.php.

```
// If the expanded menu is set, output the close button.
if ( has_nav_menu( 'expanded' ) ) {
}
```

andersnoren:
@yannickiki Good catch! I've submitted a new PR with a fix.

PR #636, opened 2019-09-28
Remove theme option to show / hide search

karmatosed:
This removes the search option and has search show all time. I 'think' I got every spot where this was but really would appreicate a code review, thanks. 

Looping @andersnoren in for design decision here as just a suggestion


ianbelanger79:
We decided to leave the option in.

Issue #633, opened 2019-09-28
Ensure valid HTML 


afercia:
I'd like to kindly remind Bundled Themes aims to output valid and semantic HTML. The HTML should be validated against https://validator.w3.org/nu/. See the core HTML Coding Standards which apply also to bundled themes: https://make.wordpress.org/core/handbook/best-practices/coding-standards/html/

Quickly checking the current output _of a single post page_ on current master 18d57a66ced1d7dae0c32b537622c124d78eb404 I spotted a few errors.

Please notice _there may be more errors_ depending on the used templates etc. and valid HTML should be checked across all the output before release.

1
**Error: Element `div` not allowed as child of element `button` in this context.** 

4 occurrences. This applies to the toggle buttons which contain a `<div class="toggle-inner">`

source:

https://github.com/WordPress/twentytwenty/blob/18d57a66ced1d7dae0c32b537622c124d78eb404/header.php#L49-L56

2
**Error: Element `div` is missing one or more of the following attributes: `role`.**

This relates to the search and menu "modal" which currently have an `aria-expanded="false"` attribute set on their `<div>` element. Since `<div>` elements aren't natively interactive, the validator throws an error.

However, `aria-expanded` should be removed from the `<div>` and used on the toggle buttons instead. I think this is already covered in another issue.

3
**Error: Element `div` not allowed as child of element `h2` in this context.**

This relates to the author h2 heading which currently contains a `<div>`. Source:

https://github.com/WordPress/twentytwenty/blob/18d57a66ced1d7dae0c32b537622c124d78eb404/template-parts/entry-author-bio.php#L12-L15

pattonwebz:
> However, `aria-expanded` should be removed from the `<div>` and used on the toggle buttons instead. I think this is already covered in another issue.

I can't find the issue that is about this. Anyone know which one it might be?

afercia:
It was mentioned here https://github.com/WordPress/twentytwenty/issues/68 (now closed)

PR #644, opened 2019-09-29
Fix CSS parse error

pattonwebz:
Ensures that instead of this:
`.comment-reply-link { color: #f5efe0; } { background: #f5efe0; }body,.entry-title a { color: #000000; }`
This is output instead:
`.comment-reply-link { color: #f5efe0; }body,.entry-title a { color: #000000; }`

I added 2 places where it is checking that there are actually `$elements` and `$selector` to generate rules for. The check at `! $selector` is partially redundant.

This is related to HTML validation done when looking into #633


nielslange:
Fixes #644

Issue #645, opened 2019-09-29
Use elements inside buttons which are not 'Interactive content'


pattonwebz:
The toggle buttons in the header all contain `<div>`s which are considered as 'Interactive content'. Buttons should not contain any interactive content:

- https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element
- https://www.w3.org/TR/2012/WD-html5-20120329/content-models.html#interactive-content

One easy solution could be to swap the `<div>` tags in the header `<button>`s to be `<span>` tags instead.

Related to the HTML validation issue outlined in #633

PR #646, opened 2019-09-29
Swap `<div>`s inside of `<button>`s to be `<span>`s

pattonwebz:
This PR swaps the `<div>` tags in the header `<button>`s to be `<span>` tags instead.

- https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element
- https://www.w3.org/TR/2012/WD-html5-20120329/content-models.html#interactive-content

Related to the HTML validation issue outlined in #633


ianbelanger79:
Fixes #645

PR #656, opened 2019-09-29
use `esc_html__` instead of `esc_html`

itowhid06:


andersnoren:
@itowhid06 Merged, thanks! What's your WordPress.org username?

itowhid06:
@andersnoren,my WordPress.org username is `itowhid06`. Thanks :)

ocean90:
@andersnoren This needs to be reverted, see #176 for background.

andersnoren:
@ocean90 I'll create a new issue and PR for changing it to `__()`. Before this PR, the string wasn't translatable.

PR #664, opened 2019-09-29
Update LICENSE

nielslange:
In comparison to https://github.com/WordPress/twentynineteen/ the license on https://github.com/WordPress/twentytwenty/ still doesn't show in the repo overview. Therefore, I copied the LICENSE from https://github.com/WordPress/twentynineteen/blob/master/LICENSE to https://github.com/WordPress/twentytwenty/blob/master/LICENSE.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,46 +1,7 @@
-WordPress - Web publishing software
-
-Copyright 2011-2019 by the contributors
-
-This program is free software; you can redistribute it and/or modify
-it under the terms of the GNU General Public License as published by
-the Free Software Foundation; either version 2 of the License, or
-(at your option) any later version.
-
-This program is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


carolinan:
Why was all the WordPress specific information removed?

ntwb:
image

It shows up for me and as @carolinan pointed out the WordPress specific information should not be removed from the `LICENSE`

nielslange:
> Why was all the WordPress specific information removed?

@carolinan I copied the license 1:1 from the Twenty Nineteen theme, which shows the license type in the repo overview. My assumption was that by using the identical file that the license type would also show for the Twenty Twenty repo.

**Twenty Nineteen:**



**Twenty Twenty:**



While Twenty Twenty shows `View license`, Twenty Nineteen shows ` GPL-2.0`.

cc: @ntwb

ntwb:
Gutenberg is the same with just "View License", this is because GitHub can detect what license is in use, but as WordPress uses an amended GPL license GitHub doesn't detect that, but we should *not* remove the specific WordPress amendments/additions of the license, if anything the Twenty Nineteen license should probably be updated

nielslange:
Thanks for your explanation, @ntwb, and consider this issue as done! 😀

PR #665, opened 2019-09-29
#633 Valid html tags fixes

### LINKED ISSUE #633: Ensure valid HTML

amolv:
#633 Valid html tags fixes


carolinan:
Hi, there is already an open pull request for this.

amolv:
PR - looks not useful, hence closing.

Issue #585, opened 2019-09-27
Small jump when opening the modals (menu & search) if the page if a bit scrolled


yannickiki:
The title is self explanatory, see the video below to see it in action. This is a small bug, but when it happens I find it a bit annoying because the scroll to top is not the expected behavior.

It wasn't happening in the theme's initial commit.

![jump](https://user-images.githubusercontent.com/54663751/65791142-668b6700-e161-11e9-9191-9bb00129d907.gif)

carolinan:
What browser are you using? Are you logged in or not? I can't see the menu bar so I'm guessing not.

yannickiki:
@carolinan on every browser (Chrome, Safari, Firefox) on Mac & Windows. I was logged in for the screencap with the admin bar disabled. It also happens for non logged in visitors.

andersnoren:
Hi @yannickiki! The scroll to the top is deliberate, as it ensures that the header and menu overlap correctly on mobile and tablet, when the header and the mobile menu are visible at the same time. Without the scroll to the top, you could end up with this:

<img width="381" alt="image" src="https://user-images.githubusercontent.com/8181091/65794667-e9fc8680-e168-11e9-9710-7a71b33c309c.png">

There could be a case made that this behavior isn't needed on desktop, though, since the header and modal menu doesn't have the same relationship there.

acosmin:
We could probably detect if we're on a mobile device and remove the jump 👍 I can do that if you guys want it.

yannickiki:
Hi @acosmin  As https://github.com/WordPress/twentytwenty/pull/628 has been merged, it seems that:

> The scroll to the top is deliberate, as it ensures that the header and menu overlap correctly on mobile and tablet, when the header and the mobile menu are visible at the same time. 

is not needed anymore as the header isn't visible when the menu modal is opened.

To avoid the jump, some of the code from the initial commit (where there was no jump) might be useful.

andersnoren:
@yannickiki @acosmin Yes, as the header is obscured on all screen sizes now, the scroll to top when a modal is displayed can be removed entirely.

acosmin:
@andersnoren I did remove the scroll to top 😄 but I had to do some change to cover template... otherwise it wouldn't work in the current setup.

PR #667, opened 2019-09-30
Fix - Scroll to top removed when modal is opened

acosmin:
Fix for #585

* scroll to top removed
* modal will remain on the same position
* For cover template I had to move `.screen-height` on the `div.cover-header` element + change a few styles. Otherwise the cover would jump top/down and its height is wrong.
* works with the admin bar.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,6 @@
 		"browser": true
 	},
 	"rules": {
-		"comma-dangle": [2, "never"],
+		"comma-dangle": [2, "never"]
 	}
 }
--- a/404.php
+++ b/404.php
@@ -12,7 +12,7 @@ get_header();
 
 <main id="site-content" role="main">
 
-	<div class="section-inner thin">
+	<div class="section-inner thin error404-content">
 
 		<h1 class="entry-title"><?php _e( 'Page Not Found', 'twentytwenty' ); // phpcs:ignore WordPress.Security.EscapeOutput.UnsafePrintingFunction -- core trusts translations ?></h1>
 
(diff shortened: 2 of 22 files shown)


andersnoren:
Using `position: absolute` to position the `.cover-header-inner-wrapper` makes it possible for the contents of the cover header to overflow at the top, so that's a no go, unfortunately :)

image

If it's only the Cover Template that has issues when we remove the scroll to the top, is it possible to only scroll to the top on the Cover Template?

acosmin:
@andersnoren it is possible. You want to try that out :)?

andersnoren:
@acosmin Let's!

acosmin:
@andersnoren done

yannickiki:
@acosmin @andersnoren  There is a problem when the cover is taller than 100 vh, the overlay is too short.
cover

Adding this CSS fixes it.
```
.cover-header-inner-wrapper {
 position: relative;
}
```

@acosmin I've tested your code and I've removed the part that scroll to the top on cover template and I can't see any issues. How can I replicate your issues with the cover template?

acosmin:
@andersnoren @yannickiki I think I got it to work with cover template 😄

yannickiki:
@acosmin I don't understand the need of the following lines. 

https://github.com/WordPress/twentytwenty/blob/73944794795b56d5ba4983c3a4b876fa67716109/assets/js/index.js#L188-L192

https://github.com/WordPress/twentytwenty/blob/73944794795b56d5ba4983c3a4b876fa67716109/assets/js/index.js#L224-L228

I don't see the connection between the cover template banner overlay and the menu modal.

I've tested without the lines above, but with the addition of 
`.cover-header-inner-wrapper { position: relative;}`that fixes the [cover overlay height issues](https://github.com/WordPress/twentytwenty/pull/667#issuecomment-536515845). I don't see any bugs. Am I missing something ? 🙂

acosmin:
Sometimes the simplest answer is the right one 😄  you were right @yannickiki 🎢

PR #680, opened 2019-09-30
Updated the twentytwenty_edit_post_link to use previous theme convention

garrett-eclipse:
Updated the twentytwenty_edit_post_link to use the convention from twentynineteen and previous themes to have the post title available in screen reader text.

This changes the link from Edit This to Edit, but screen readers will get Edit Post Name.


garrett-eclipse:
Thanks @andersnoren & @carolinan for reviewing and merging this. All the best

Issue #679, opened 2019-09-30
Header Search place Close far-right same as Search


garrett-eclipse:
Hello,

Very minor UX issue with the search when open the close is slightly displaced off to the left so you have to move cursor to close. Would be nice if it was right aligned so you can open/close without moving your cursor.

![3350818592ff1db5f1ad8b0f6be1768b](https://user-images.githubusercontent.com/8726005/65917505-54414100-e38c-11e9-93b3-62341ad137fb.gif)

Thanks

andersnoren:
Hi @garrett-eclipse,

Thanks for creating the issue! The cross is positioned to be aligned with the right side of the outermost wrapper of the theme (1200px wide). I think the search field text would look unbalanced if the cross was positioned all the way to the right in search overlay, especially on larger screens.

I do agree though that the user should be able to close the search overlay by clicking in the same place as they clicked to open it. One way to solve this would be to add a pseudo element on the `.search-toggle` button element, and position that pseudo element to stretch all the way to the right edge of the screen in the overlay, like so:

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/8181091/65963014-29bdb980-e45a-11e9-996c-d9192a0d143a.png">

The CSS would need to be modified a bit on mobile, to make the positioning work there.

garrett-eclipse:
Thanks @andersnoren that makes sense. And on mobile not sure we'll be able to align as the search flips to be on the left hand side.

glauberm:
@andersnoren @garrett-eclipse Won't be weird to click on a white space to close the search form? I believe a better solution would be to make the search form follow the `max-width: 168rem` from `.header-inner` instead of the `max-width: 120rem` from `.section-inner`, no?

andersnoren:
@glauberm After testing your suggestion on my local, I agree with you. I saw @amolv started on a PR using this solution, so I'll tell him to go ahead with it.

glauberm:
@andersnoren That's a shame. I was looking forward to implement this solution. Anyway, I'm glad that it was accepted. :)

amolv:
I created PR before this discussion.

andersnoren:
@amolv No worries! I've left you a comment on the PR.

amolv:
> @amolv No worries! I've left you a comment on the PR.

Thank you, PR updated.

ianbelanger79:
Fixed in #687

PR #687, opened 2019-10-01
#679 Header Search icons place Close far-right same as Search fix

amolv:


style.css
```diff
@@ -919,7 +919,7 @@ section {
 .section-inner {
 	margin-left: auto;
 	margin-right: auto;
-	max-width: 120rem;
```
andersnoren: This helper class is used for other elements than the search modal, so you can change this line back to `120rem`.

amolv: Sorry, its back to 120rem now.

andersnoren:
@amolv What's your WordPress.org username?

amolv:
> @amolv What's your WordPress.org username?

amolv - https://profiles.wordpress.org/amolv/

Issue #598, opened 2019-09-27
Unexpected click outline


karmatosed:
I was testing the menu and got this unexpected image:

![image](https://user-images.githubusercontent.com/253067/65799381-e8808d80-e16b-11e9-9364-1ffde2ab9c39.png)

After digging I also get on desktop:

![image](https://user-images.githubusercontent.com/253067/65799405-f59d7c80-e16b-11e9-9956-3307f7e73818.png)

karmatosed:
It seems to not just be on menus. I found it also on search:

![image](https://user-images.githubusercontent.com/253067/65799623-7e1c1d00-e16c-11e9-86aa-9a6ea966abf4.png)

afercia:
This appears to be the browser's native focus indication.

Are you testing with the Opera browser? (I see the outline is yellow, which is only used by Opera I think 🙂 )

Not a bug _per se_  I'd say but maybe the focus indication could be improved. Two consderations:

1
The particular shape of the focus on Search and Menu with that sort of protruding rectangle is given by the `width: 10rem;` applied to the button text:

Opera:

<img width="501" alt="Screenshot 2019-09-28 at 12 05 25" src="https://user-images.githubusercontent.com/1682452/65814974-7058ad00-e1e9-11e9-9992-20264d8caa20.png">
<img width="492" alt="Screenshot 2019-09-28 at 12 05 34" src="https://user-images.githubusercontent.com/1682452/65814975-7058ad00-e1e9-11e9-8c81-7409a5a4c0cd.png">

Chrome:

<img width="596" alt="Screenshot 2019-09-28 at 12 08 23" src="https://user-images.githubusercontent.com/1682452/65814986-8f573f00-e1e9-11e9-8fbe-303eb0791d22.png">

I'd tend to think `width: 10rem;` is a "magic number" that should be avoided with some refactoring of the CSS.

2
Worth reminding the browsers native focus style differs across browsers. In Firefox the native style is very light and a custom style would be recommended.

Firefox:

<img width="729" alt="Screenshot 2019-09-28 at 12 09 49" src="https://user-images.githubusercontent.com/1682452/65815008-dd6c4280-e1e9-11e9-9ab3-df7898e6a35e.png">
<img width="707" alt="Screenshot 2019-09-28 at 12 09 59" src="https://user-images.githubusercontent.com/1682452/65815009-dd6c4280-e1e9-11e9-8f6f-765a33b7a3e9.png">

karmatosed:
[...]

PR #691, opened 2019-10-01
Fix ugly outline on header buttons on Google Chrome

audrasjb:
Hello,

Here is a proposal to fix the outline issue on Google Chrome.
First, I removed the default outline, using `outline: none`.
Then, I used an inset box shadow to replace the default outline.

This could probably be improved, but to me, the result looks better than the current outline.
See screenshots below.

Fixes: #598 

Cheers, 
Jb

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5318,8 +5318,7 @@
 				"ansi-regex": {
 					"version": "2.1.1",
 					"bundled": true,
-					"dev": true,
-					"optional": true
+					"dev": true
 				},
 				"aproba": {
 					"version": "1.2.0",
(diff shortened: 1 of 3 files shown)


audrasjb:
Before:
Capture d’écran 2019-10-01 à 14 00 34

After:
Capture d’écran 2019-10-01 à 13 59 36

carolinan:
> Oops, sorry, my bad :\
I am bit confused, why are you apologizing? :)

But I think the focus outline needs a design decision first. I think it needs to be consistent, and this shadow will stand out from other elements with focus.

PR #693, opened 2019-10-01
Add missing roles to <nav> elements

audrasjb:
Adds missing `navigation` roles to `<nav>` elements.

Fixes: #462


carolinan:
;P What happened to the header banner role?
https://github.com/WordPress/twentytwenty/blob/master/header.php#L33

PR #694, opened 2019-10-01
Fix: Search query term is not visible in the search input field

maheshwaghmare:
The search query term is not visible in the search input field.

If we search any tern then it is visible in the search input.

E.g. If we search `testing` then it is visible in the search box.

See screenshot of theme Twenty Nineteen:
Theme Twenty Nineteen

But, It is not visible in TwentyTwenty.

See screenshot - (Before Fix Theme TwentyTwenty):
Before Fix Theme Twenty Twenty

---

The issue is in file `twentytwenty/searchform.php` line 24 the code is:

```
value="<?php get_search_query(); ?>"
```

This should be:
```
value="<?php echo get_search_query(); ?>"
```

After issue fix screenshot (After Fix Theme TwentyTwenty):
After Fix Theme Twenty Twenty


nielslange:
Nice catch and thanks for your first contribution, @maheshwaghmare! 😀

pattonwebz:
@maheshwaghmare if this is your first contribution could you add your name to the contributors file as well in this PR? Alternatively share your wordpress.org username here and I'll add your name to the list.

maheshwaghmare:
My WordPress username is [@mahesh901122](https://profiles.wordpress.org/mahesh901122/).

carolinan:
Why should the search term be visible in the field?  I mean it is not like we are going to use the same search term again?

joyously:
It makes it easier for the user if there was a small typo in the search term.
I think it's a good idea.

utsavtilava:
Thanks @pattonwebz.

My WordPress username is @utsav72640.

Issue #697, opened 2019-10-01
Padding difference in front end and back end for Table block in Gutenberg


Swapnildhanrale:
As shown in the screenshots below.
# In 2020 Theme
Back end Screenshot Url - https://share.getcloudapp.com/E0uPWZGX 
Front end Screenshot Url - https://share.getcloudapp.com/bLup656N

# In 2019 Theme
Back end Screenshot Url - https://share.getcloudapp.com/wbue1Qeq
Front end Screenshot Url - https://share.getcloudapp.com/p9uzdQjd

ianbelanger79:
Duplicate of #320

PR #699, opened 2019-10-01
Padding difference in front end and back end for Table block in Gutenberg.

Swapnildhanrale:
The theme support for the Block Style's is not added.

```
// Add support for Block Styles.
add_theme_support( 'wp-block-styles' );
```

And `12px` padding added to the table block cell content as below:

```
.editor-styles-wrapper .wp-block-table .wp-block-table__cell-content {
      padding: 12px;
}
```

By default, the block editor added `0.5em` to the table cell content which is overridden by above CSS.

For that, the table does not look the same in block editor and on the front end.

Fixes: #697


ianbelanger79:
Thanks for submitting your patch @Swapnildhanrale, however the theme contains it's own block styles, thus the core block styles are not enqueued. You can update your patch to remove the core block styles and we will test again. Thanks

Swapnildhanrale:
Hello @andersnoren I have just removed the block style support you can please check and merge my PR

ianbelanger79:
This does not fix the table issues at the moment. Padding is still off between back and front end. Also when using the `Fixed width table cells option` and Wide or Full width, the table is misaligned and is not wide or full width.

ianbelanger79:
Closing because the PR is out-of-date and the issues have been fixed in another PR

Issue #459, opened 2019-09-22
Left aligned text is cut off in full width blocks


carolinan:
Left aligned text inside full width blocks are missing a left margin or padding, so the text is cut off:

![half](https://user-images.githubusercontent.com/7422055/65385900-a2bf6180-dd34-11e9-87d5-7334e49c9d69.png)

If I add a background color to the group block, there is a large left padding:

![half2](https://user-images.githubusercontent.com/7422055/65385950-6f310700-dd35-11e9-8231-0ffa0b9ddbce.png)

ianbelanger79:
Fixed in #707

PR #707, opened 2019-10-01
Adds padding for paragraphs in full-width columns block

luminuu:
The media queries used here reflect the media queries for the columns block from core. The changes exclusively work for paragraphs which are direct child elements of a single `wp-block-column`, where the `wp-block-columns` wrapper is set as `alignfull`.

@media ( max-width: 599px ) : paragraphs get the same width as the section-inner class.

@media ( min-width: 600px ) : there's a maximum of two columns displayed, so all columns need to get either padding-left or padding-right via :nth-child(odd) or :nth-child(even).

@media ( min-width: 782px ) : all columns are displayed in one row, so only the first and last child items need some padding.


nielslange:
@andersnoren Would you mind taking a look at change as proposed in https://github.com/WordPress/twentytwenty/pull/707#issue-323401431?

luminuu:
I have updated the CSS to match the requirements in #459 in my latest commit: f1d5376 

I added support for the group block, added a `:not(.has-background-color)` pseudoclass to avoid double paddings and removed the first `>` to apply this to all nested elements.

luminuu:
@andersnoren Thanks for the feedback! I updated the targeting of the group block as you suggested.

ianbelanger79:
Hi @luminuu, what is your WordPress username? Please let me know asap, thanks

luminuu:
Hey @ianbelanger79, it's the same as my Github username: https://profiles.wordpress.org/luminuu

nielslange:
> Hey @ianbelanger79, it's the same as my Github username: https://profiles.wordpress.org/luminuu

@ianbelanger79 I've added @luminuu to https://github.com/WordPress/twentytwenty/blob/master/CONTRIBUTORS.md in https://github.com/WordPress/twentytwenty/pull/811 and added you as a reviewer. 😀

PR #709, opened 2019-10-01
RSS Widget: hide full link instead of just the related image

audrasjb:
Instead of removing the RSS image, it would be better to remove the full link. Without that, the link will be vocalized as an empty link by assistive technologies. 

Worth noting I had to use `:first-of-type` pseudo selector since there is no specific class on this element.

Cheers,
Jb


andersnoren:
@audrasjb Thanks, merged!

PR #710, opened 2019-10-01
Use core's function for single post navigation

anace:
This PR solves issue #300, as in PR #314, PR #350 and PR #439. The markup is simpler and accessibility ready.


briceduclos:
@anace Can you please make the requested changes? I would like to submit a PR for #793 which depends on this PR.

joyously:
Are all the classes with 'pagination' in the name actually for post navigation? Or are some for pagination (like on archive pages)?
It should definitely be consistent with the template part file.

Issue #147, opened 2019-09-09
Unnecessary use of escaping functions


acosmin:
https://github.com/WordPress/twentytwenty/blob/c3fd9652e30fb489d83c493b4451ac361f81d75d/functions.php#L301-L303

https://github.com/WordPress/twentytwenty/blob/35b5d240fcbf283d753f95b9b107be4200fdfee4/footer.php#L17

https://github.com/WordPress/twentytwenty/blob/35b5d240fcbf283d753f95b9b107be4200fdfee4/parts/modal-search.php#L7

For `$cover_header_classes`:
https://github.com/WordPress/twentytwenty/blob/35b5d240fcbf283d753f95b9b107be4200fdfee4/content.php#L45

https://github.com/WordPress/twentytwenty/blob/35b5d240fcbf283d753f95b9b107be4200fdfee4/content.php#L74

https://github.com/WordPress/twentytwenty/blob/35b5d240fcbf283d753f95b9b107be4200fdfee4/content.php#L142

https://github.com/WordPress/twentytwenty/blob/35b5d240fcbf283d753f95b9b107be4200fdfee4/comments.php#L57

https://github.com/WordPress/twentytwenty/blob/35b5d240fcbf283d753f95b9b107be4200fdfee4/parts/classes/class-comment-walker.php#L26

ianbelanger79:
Can we get a PR for this one @pattonwebz? I would like to close it out. Thanks

acosmin:
I could probably take a crack at it :)

PR #719, opened 2019-10-02
Fix - Unnecessary use of escaping functions

acosmin:
Fix for #147 

I also wrapped a few `get_the_title()` instances with `esc_html`, in places where adding `em` or `strong` tags would look bad (like the read more link)`

@ianbelanger79 I think I got them all 😄


andersnoren:
@acosmin Can you sneak in this one as well? :) https://github.com/WordPress/twentytwenty/pull/656#issuecomment-537432633

acosmin:
@andersnoren no problem :) did it.

Issue #720, opened 2019-10-02
Inline icon label when no "Desktop Horizontal Menu" is set


yannickiki:
Currently when there is no "Desktop Horizontal Menu" set, the labels for the Menu and Search icons are stacked. According to the Figma mockups, the labels of Menu and Search icons should be inline with a bigger font-size.

**Before:**

![stack](https://user-images.githubusercontent.com/54663751/66050151-ef711c80-e52c-11e9-9a40-a775394fba26.jpg)

**After:**

![inline](https://user-images.githubusercontent.com/54663751/66050152-f0a24980-e52c-11e9-992b-e92d200ec623.jpg)

@andersnoren could you please confirm this? 🙂

andersnoren:
@yannickiki Yeah, the original plan was to use that toggle layout on desktop when the "Horizontal Desktop Menu" is not set. Given that the header is already pretty complex, we decided to leave this one as a "nice-to-have" in case there's time. Since the expanded menu will only be used if opted into, and the primary menu will not be used only if opted out of, the number of users that will only have the expanded menu set is probably pretty low.

But if someone feels inclined to implement this update on > 1000px screens, it's up for grabs. 🙂 It would need to be done pretty quickly though, and preferably only with CSS changes.

glauberm:
I'll try to work on a PR addressing this today.

glauberm:
#723 @andersnoren I understood from what you said that I should keep the changes to a minimum, so I didn't follow the after picture styles very thoroughly, ignoring the font size and color, and only guessing the spacing between the icon and the label. Let me know if I should apply these styles.

glauberm:
Also, working on this issue I found a missing bracket on [line 4317 of `style-rtl.css`](https://github.com/WordPress/twentytwenty/blame/a7fab672077f31bd232318ee2d3cd3cce1b8e15f/style-rtl.css#L4317). Weirdly enough, [the blamed PR](https://github.com/WordPress/twentytwenty/pull/576/files) has the bracket... Anyway, let me also know how you wanna handle this.

PR #723, opened 2019-10-02
#720 Inline menu labels and icons on desktop (>=1000px)

glauberm:


andersnoren:
@glauberm Thanks for taking a look at this. It looks like me bringing this branch up to date caused the changes you made in the "inlining menu labels" commit to vanish. Sorry about that. Do you think you can update the PR with the changes in that commit again?

I applied the changes in that commit in my local, and I noticed a couple of things:

- It changes the toggles to always display the text next to the icon. That layout should only be used when a horizontal menu is **not** set. When a horizontal menu is set, the old layout (with text beneath icons) should be used.
- When the toggle text is displayed next to the icon, the text should have a font-size of 15px, a right padding of 20px and the main body color, rather than the light gray one.

glauberm:
No problem, @andersnoren . I'll work on those changes ASAP.

glauberm:
@andersnoren The CI check failed but I'm not sure why. Do you think opening a fresh new PR would help with this issue?

ianbelanger79:
@glauberm this is why your PR failed

```
style.css
 5441:5  ✖  Expected indentation of 2 tabs   indentation
```

Fix that and it should pass

glauberm:
Fixed @ianbelanger79.

style.css
```diff
@@ -5209,7 +5209,6 @@ a.to-the-top > * {
 		position: static;
 		padding-right: 20px;
 		font-size: 15px;
-		color: #000;
```
glauberm: @ianbelanger79 Doesn't removing this line undo the following requirement pointed out by @andersnoren?

> * When the toggle text is displayed next to the icon, the text should have a font-size of 15px, a right padding of 20px **and the main body color, rather than the light gray one**.

ianbelanger79: It may remove that requirement, but having a set color causes issues when the Header Background color is set to `#000` or any dark color. The toggle text is no longer visible. @andersnoren do you agree?

andersnoren: @ianbelanger79 @glauberm Setting it to `color: inherit` here instead will make it inherit the color from the toggle element, which is updated with the color settings.

ianbelanger79:
@glauberm What is you WP username? Please let me know asap, thanks

PR #733, opened 2019-10-03
I18N: Reduce use of context functions and improve translator comments

### LINKED ISSUE #732: Don't explain placeholders in context
There are a few strings which incorrectly use the `_x()` or `_nx()` functions to explain the placeholders even through they are already explained in the translator comment.

Examples:

https://github.com/WordPress/twentytwenty/blob/303e31c8d979bd24b46c874219211617a7848d33/footer.php#L46-L47
Better:
```
/* translators: %s: HTML character for an arrow. */
printf( __( 'To the top %s', 'twentytwenty' ), '<span class="arrow">&uarr;</span>' );
```

---

https://github.com/WordPress/twentytwenty/blob/303e31c8d979bd24b46c874219211617a7848d33/footer.php#L26
Better:
```
/* translators: Copyright date format, see https://secure.php.net/date */
_x( 'Y', 'copyright date format', 'twentytwenty' )
```

ocean90:
Fixes #732.

### DIFF
--- a/404.php
+++ b/404.php
@@ -21,7 +21,7 @@ get_header();
 		<?php
 		get_search_form(
 			array(
-				'label' => _x( '404 not found', 'Label', 'twentytwenty' ),
+				'label' => __( '404 not found', 'twentytwenty' ),
 			)
 		);
 		?>
--- a/footer.php
+++ b/footer.php
@@ -43,14 +43,14 @@
 					<a class="to-the-top" href="#site-header">
 						<span class="to-the-top-long">
 							<?php
-							// Translators: %s = HTML character for an arrow.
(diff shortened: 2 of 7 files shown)


ocean90:
@nielslange Marking the PR as a draft was intentionally since it's not finished yet…

nielslange:
@ocean90 My bad. I pushed the wrong button accidentally! 🤭 Won't happen again! Is there any way I can revert that?

ocean90:
@nielslange Doesn't look like there's an option for reverting back to draft yet. But the PR should be good now.

Related to https://github.com/WordPress/twentytwenty/pull/733/commits/b6c08129d7b3b4347430d36eeb223d59b4489366 is #734.

nielslange:
> @nielslange Doesn't look like there's an option for reverting back to draft yet. But the PR should be good now.

@ocean90 I did a quick research and it seems this is a missing feature. Many folks requested it, but it's indeed not possible to change a PR that's ready back to a draft.

PR #737, opened 2019-10-03
Removed underline on mouseover of Post Comment button

rajilesh:
Removed underline from Post Comment button, so that all button style will be consistent across the site.

### DIFF
--- a/style.css
+++ b/style.css
@@ -3889,6 +3889,10 @@ div.comment:first-of-type {
 	display: block;
 }
 
+.comment-respond #submit:hover {
+	text-decoration: none;
+}
+
 .comment-respond .comments-closed {
 	text-align: center;
 }


andersnoren:
Hi @rajilesh, thanks for creating the PR. All buttons have (and should have) `text-decoration: underline` on hover, so I'm closing this PR. Let me know if I've misunderstood something and the PR should be reopened.

rajilesh:
@andersnoren I think underlines are good for text links but it is not needed for buttons in my opinion as it will not look like button. We can have background colour change on hover.

For example: Google homepage buttons

Screenshot 2019-10-04 at 12 41 13 PM

rajilesh:
@andersnoren on homepage search button also doesn't have text-decoration.

nielslange:
@rajilesh The search button on the homepage looks like this to me on `:hover`:

Screenshot 2019-10-04 at 14 24 53

And according to https://github.com/WordPress/twentytwenty/blob/ed551762647d2e5ad975198b02a709d79aabd188/style.css#L815-L832 it is using `text-decoration: underline;` on `:hover`. I guess the problem is that some buttons do not look like buttons, thus, that's something that could be improved. 

@andersnoren Would it be an option to introduce a different `:hover` style for buttons that look as follows:

Screenshot 2019-10-04 at 14 27 43

rajilesh:
I mean this search button.
Screenshot 2019-10-04 at 1 31 07 PM

rajilesh:
@nielslange sure, we can have different style on hover. but the underline for the button is not needed in any case. I think.

nielslange:
Thanks for clarifying this, @rajilesh. Personally, I would also use a different `:hover` style for these buttons, but it's @andersnoren's final call as design lead of the Twenty Twenty team. Let's see how he thinks about this.

rajilesh:
Yup.

PR #741, opened 2019-10-03
Featured Images: do not display on posts using the Image Post Format

jeherve:
Although Twenty Twenty itself does not declare support for Post Formats, one may have existing posts using Post Formats.

If that's the case, when viewing single posts you will see a quite large featured image followed by the post content, including an image that is very likely to be the same as the featured image. This PR avoids displaying the featured image in such scenarios.

I feel this is an important addition to Twenty Twenty as the Featured Images (especially in a portrait orientation) can take quite a lot of space.
I've limited that change to the single post view, as folks will typically not see the post image on index views (they will see an excerpt instead, which does not show the post image).

WordPress.org username: `jeherve`

### DIFF
--- a/template-parts/content.php
+++ b/template-parts/content.php
@@ -19,7 +19,10 @@
 
 	get_template_part( 'template-parts/entry-header' );
 
-	if ( ! is_search() ) {
+	if (
+		! is_search()
+		&& ! ( is_singular() && has_post_format( 'image' ) )
+	) {
 		get_template_part( 'template-parts/featured-image' );
 	}
 


pattonwebz:
Did you test if it would be the case that posts would show doubled images here? 

If the theme does not handle output of the images before the content then I don't know how it would double them up. I think if this is a problem for people it is quite an edge case.

ianbelanger79:
I agree with @pattonwebz, this is most likely an edge case, thus we will not be merging this PR. Thanks for submitting.

jeherve:
> Did you test if it would be the case that posts would show doubled images here?

Right, that's the bug, hence why I opened this PR. :) Twenty Twenty shows doubled images in that case: Featured Image + Post Image.

> I think if this is a problem for people it is quite an edge case.

That's fair enough. I suppose Post Formats are a bit of an edge case nowadays.

Thanks for the review.

PR #742, opened 2019-10-03
Replace uniqid() with wp_unique_id()

### LINKED ISSUE #735: Replace uniqid() with wp_unique_id()
See https://core.trac.wordpress.org/ticket/44883 for background.

felipeelia:
Fix #735


joyously:
What was wrong with using uniquid()?

PR #746, opened 2019-10-03
Replace wp_unique_id with twentytwenty_unique_id to retain < 5.0.3 compat

### LINKED ISSUE #743: Add back-compat.php?
It looks like the theme requires at least 5.0.3 after #742. Should a [back-compat.php](https://github.com/WordPress/WordPress/blob/5.2-branch/wp-content/themes/twentynineteen/inc/back-compat.php) file be added to ensure the theme can't be enabled on older WordPress versions?

### LINKED ISSUE #745: Replace wp_unique_id to retain < 5.0 compatibility
The `wp_unique_id` call used in searchform.php needs to be replaced in order to retain compatibility with WordPress versions older than 5.0.3. Twenty Seventeen solves this by including and calling its own function that uses `wp_unique_id` when available. More info: https://core.trac.wordpress.org/ticket/44883

andersnoren:
Replaces the `wp_unique_id` call in `searchform.php` with a `twentytwenty_unique_id` call, which uses `wp_unique_id` if the function exists. Fixes compatibility issue with versions older than 5.0.3.

Fixes #745.
Closes #743.


carolinan:
> ~@andersnoren The code doesn't reflect the title of this issue. It seems that this the identical code that @ocean90 changed with [ed55176](https://github.com/WordPress/twentytwenty/commit/ed551762647d2e5ad975198b02a709d79aabd188) already.~
> 
> Please ignore my previous comment. I saw the code of [ed55176](https://github.com/WordPress/twentytwenty/commit/ed551762647d2e5ad975198b02a709d79aabd188) initially.

Sorry this was because I merged them in the wrong order >. <

nielslange:
> Sorry this was because I merged them in the wrong order >. <

No problem, @carolinan, and thanks for clarifying that. I just had a _"Uh, what just happened?!?"_ moment. 😄

carolinan:
I give up, I should not have tried this from my mobile 🙄 If its not fixed before I get home I will solve it then.

acosmin:
I also don't think this should be escaped anymore: https://github.com/WordPress/twentytwenty/blob/ed551762647d2e5ad975198b02a709d79aabd188/searchform.php#L22

Issue #722, opened 2019-10-02
Social Menu icons jump 1px on hover


ianbelanger79:
When you mouseover a social menu icon, it jumps 1px to the right and the icon gets blurry.

Before Hover
![fb-icon-no-hover1](https://user-images.githubusercontent.com/34601694/66071502-4113aa00-e521-11e9-8f86-02d3ee9092ba.PNG)

On Hover
![fb-icon-hover](https://user-images.githubusercontent.com/34601694/66071413-19244680-e521-11e9-8d3b-222843d72915.PNG)

glauberm:
Which browser were you testing on? I think this is a issue on how certain browsers handle the `transform: scale` transition. Not sure how to fix it.

andersnoren:
@ianbelanger Can you check if this still happens in #717?

ianbelanger79:
#717 did not fix it @andersnoren. I tested on Windows in Firefox and Chrome and see the issue in both. Surprisingly, IE11 and Edge do not have this issue.

glauberm:
I'm taking a look into that and I believe it's happening due to a mismatch between the paths size and the viewbox attribute. Soon I'll make a PR out of it.

glauberm:
@ianbelanger79 Could you verify if #756 solves the problem?

ianbelanger79:
Fixed by #933

PR #756, opened 2019-10-04
#722 Fix social icons

glauberm:
All the SVGs were a little off, so I rebuilt them. I also reordered the GitHub one, that was not following the alphabetical order.


carolinan:
What _exactly_ does "Little off" and "rebuilt" mean?

glauberm:
@carolinan They're fundamentally the same icons. I'll use the facebook one to demonstrate it. Here's the old icon with an added border to show its limits:

old-icon

It has an irregular whitespace in it (that's what I meant by "a little off"), which makes it scale irregularly. My first thought was to change its `width`, `height` and `viewport`, thinking that it would make the whitespace disappear. Unfortunately, it didn't:

failed-attempt

I've even tried to change the `x` an `y` attributes, believing that would do it, but it didn't. I'm not sure why. Anyway, I decided to follow my usual SVG workflow and open the file on Inkscape, make the canvas fit the drawing and then save it as an [optimized SVG](https://www.integraxor.com/optimize-inkscape-svg-files/) with a precision of 4 digits for the coordinates (that's what I meant by "rebuilt"). That resulted on basically the same icon, but with no whitespace at all:

new-icon

I realize that the optimal solution would be to don't change the `path` coordinates at all, but I was unable to solve the problem without doing that. Perhaps we could "rebuilt" them again with more precision... Let me now what you think.

pattonwebz:
This looks like a nice little tweak could save people from doing micro adjustments via CSS in their projects if they use the icons from the theme. That being said I think any changes to icons like this will need to wait until after a release is made as we are currently running the RC, let's see if any feedback comes back about this needed that could make us consider the misalignment a bug.

glauberm:
Fixed by #933

PR #762, opened 2019-10-05
Adjust retina logo setting

### LINKED ISSUE #752: "Retina logo" setting isn't working. Consider replacing it with a "Logo width" setting.
- Currently the "Retina logo" option isn't working. It only changes the width and height attributes of the logo  tag which doesn't make any change visually. CSS is needed to make it appear half its size.

- Consider renaming the "Retina logo" label as Retina is an Apple trademark

- Currently the goal is to scale the logo to half its uploaded size (x2). However, more and more smartphones and tablets have now higher screen density (x3 / x4). So a x2 logo wouldn't appear sharp on them. As it’s not possible to upload svg, a solution might be to replace the “Retina logo” checkbox by a “Logo width” input number.

andersnoren:
- Removes a line of CSS that overwrote the `height` and `width` attributes on the logo image, which caused the setting to not have any effect.
- Renames `retina_logo` to `high_resolution_logo`, to make it less Apple specific.
- Removes the `header-text` parameter from `add_theme_support( 'custom-logo' );`, since we include our own logic for whether to display the site title or the site logo and included style for using the site logo and site description together.

Fixes #752.


carolinan:
Works well, but the change is not previewed in the customizer, can we change the transport?

yannickiki:
(as said in #752) If the logo is higher than 6rem on mobile or 9rem on tablet and desktop (max-height for the logo image defined by the theme), the logo isn't resized proportionally. As a max-height is set for the logo, removing width: auto causes this issue.

andersnoren:
@yannickiki Good catch  – complete brain fart on my part. I'll add an alternate solution to this PR this morning.

andersnoren:
Updated with a new solution for the retina logo setting, and selective refresh for  `high_resolution_logo`.

pattonwebz:
I have tested this on my end and it seems to be working fine and also the transport works correctly again.

If the sizing issues as noted in #752 become a problem again let's revisit in a fresh issue thread.

PR #768, opened 2019-10-05
Fixed typo and updated some docblocks

itowhid06:


classes/class-twentytwenty-customize.php
```diff
@@ -420,7 +422,9 @@ public static function sanitize_select( $input, $setting ) {
 		/**
 		 * Sanitize boolean for checkbox.
 		 *
-		 * @param bool $checked Wethere or not a blox is checked.
+		 * @param bool $checked Wethere or not a box is checked.
```
ntwb: Spelling: Whether

inc/template-tags.php
```diff
@@ -26,6 +26,8 @@
  *
  * @param array   $args Arguments for displaying the site logo either as an image or text.
  * @param boolean $echo Echo or return the html.
+ *
+ * @return string $html Compiled html based on our arguments.
```
ntwb:
```suggestion
 * @return string $html Compiled HTML based on our arguments.
```

ntwb: Uppercase HTML for consistency with the other instances of this in this PR

nielslange: @itowhid06 In line 28 there's another `html` which with should be `HTML`.

PR #770, opened 2019-10-05
Add filters for the lists of icons

andersnoren:
Adds two new filters, `twentytwenty_svg_ui_icons` and `twentytwenty_svg_social_icons`, which can be used to modify the two lists of icons included in Twenty Twenty.

Also changes the default value of the `TwentyTwenty_SVG_Icons::get_svg()` color parameter from `#1A1A1B` to `#000000`, to match the front-end of the theme.


acosmin:
`$arr = apply_filters( 'twentytwenty_svg_icons', $arr, $group );` would be simpler :)

acosmin:
or: `$arr = apply_filters( "twentytwenty_svg_$group_icons", $arr );`. Also, passing `$icon` and `$color` as args would make it more filterable.

andersnoren:
@acosmin Good points – changed it to a single filter with a dynamic hook name. Also added the filter to `get_social_link_svg()`.

PR #782, opened 2019-10-06
Made the footer credits area hookable

acosmin:
This PR adds an action `twentytwenty_footer_credits` where `twentytwenty_footer_copyright` & `twentytwenty_footer_powered_by` are hooked. This is to make it simpler to modify the credits without replacing the entire file or hiding it with `display: none`


ocean90:
I don't think it needs to be that complex. I'd rather change this to a template part which a child theme can override.

`<?php get_template_part( 'template-parts/footer-credits' ); ?>`

acosmin:
I guess you're right :) my only concern is that it involves a few more steps for a simple user, like ftp, uploading files. In this way you just tell the user to add some code in functions.php and he's done.

nielslange:
> I guess you're right :) my only concern is that it involves a few more steps for a simple user, like ftp, uploading files. In this way you just tell the user to add some code in functions.php and he's done.

@acosmin And with the next theme update, the changes are gone? 😉 I would always recommend usinga child theme.

acosmin:
@nielslange yes, I forgot to mention that :)

acosmin:
I could make it simpler by just adding the `do_action` call in footer.php, without the wrapper function...

pattonwebz:
This one will need to be moved to trac and considered after WP 5.3 release. When this repo is being closed I can port this PR over to a trac ticket and create a patch file for it with the changes here.

Issue #786, opened 2019-10-07
Update text strings


carolinan:
Some text strings need to be updated before the string freeze in beta 3.
-Remove the theme name from the Theme Options customizer panel.
-Remove the word "image" from the cover template options since the image is optional.

PR #787, opened 2019-10-07
Update and remove text strings in the customizer

carolinan:
For #786 but there may be more text that needs to be updated.

### DIFF
--- a/classes/class-twentytwenty-customize.php
+++ b/classes/class-twentytwenty-customize.php
@@ -191,9 +191,9 @@ if ( ! class_exists( 'TwentyTwenty_Customize' ) ) {
 			$wp_customize->add_section(
 				'options',
 				array(
-					'title'       => __( 'Theme Options', 'twentytwenty' ),
-					'priority'    => 40,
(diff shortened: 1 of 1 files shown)


classes/class-twentytwenty-customize.php
```diff
@@ -252,13 +251,13 @@ public static function register( $wp_customize ) {
...
 					'capability'  => 'edit_theme_options',
-					'description' => __( 'Settings for the "Cover Template" page template.', 'twentytwenty' ),
+					'description' => __( 'Settings for the "Cover Template" page template. Add a featured image to use as background.', 'twentytwenty' ),
 					'priority'    => 42,
 				)
 			);
 
 			/* Overlay Fixed Background ------ */
-
+
```
nielslange: I believe this space will trigger en Travis error, @carolinan.

nielslange: I just see that you solved it in https://github.com/WordPress/twentytwenty/pull/787/commits/9f5c8db319ebdb6cc4fa544ad394191916326d83 already! 😁

PR #788, opened 2019-10-07
Remove Background image option

### LINKED ISSUE #480: Consider removing support for background-image
Using a background image can make a website less accessible. Considering all the efforts that has been made to keep the content accessible depending on the background colors chosen, I would suggest to remove the option of setting a background image in the Customizer. It would also make the Customizer options more limited. This option can be easily added using a child theme.

We did the same for Twenty Nineteen.

carolinan:
This removes the customizer section for the background image.
Fixes #480

_Removing existing sections or controls is not in accordance with the Theme Review requirements._
We _shouldn't_ add theme support for something, only to remove it.

However:
The background image is part of the theme support for custom background.
The alternative would be to remove the theme support, which would also remove the background color control.
A new custom color control would need be added and tied to the a11y color contrasts code.
Because of this, the faster, easier shortcut is to remove the background image section.


acosmin:
I agree with this one 😄 I also think we should amend the requirement and add a few exceptions in there, like this one.

carolinan:
FYI Theme Review representatives have agreed that an exception can be made.

PR #789, opened 2019-10-07
Fix: Hard to style quote and pullquote color.

jorgefilipecosta:
Currently, the 2020 theme sets a color to inherit on the quote and pull quote blocks.
The block editor comes with an easy way o register block styles and the samples we provided were similar to:
```
register_block_style(
	'core/quote',
	array(
		'name'         => 'not-fancy-quote',
		'label'        => 'Not Fancy Quote',
		'inline_style' => '.wp-block-quote.is-style-not-fancy-quote { color: blue; }',
	)
);
```
Using 2020 theme the previous sample does not works as expected because the color set is overwritten by the inherit color styles.

I think this inherit color styles are not needed as the blocks are already inheriting by default on these cases.
By removing these styles, I did not notice any change in the blocks.

Without these styles removed the code sample I provided when pasted on functions.php of the theme does not work on the editor (the color blue does not appear on the quote block when Not Fancy Quote block style is selected) with these styles remove the code sample works as expected.


ianbelanger79:
@andersnoren could you review this?

jorgefilipecosta:
Hi @andersnoren, thank you for the review! This PR was updated and now we only apply the change on quote block.

PR #794, opened 2019-10-07
Updates to RTL stylesheets

ianbelanger79:


nielslange:
@ianbelanger79 How can we ensure that these chances don’t get overwritten when running `npm run build` in the future?

ntwb:
> How can we ensure that these chances don’t get overwritten when running npm run build in the future?

If RTL files require certain things to not be overwritten then using the RTLCSS _control directives_ to ignore as required is the solution to that, see:
https://rtlcss.com/learn/usage-guide/control-directives/

nielslange:
Thanks, @ntwb. I'll create another PR to ensure that @ianbelanger79's changes will not be overwritten in the future.

nielslange:
@ianbelanger79 I might have misunderstood your PR. You haven't added custom styles to the RTL styles but you've synced them with the LTR sheets, correct?

ianbelanger79:
That is correct @nielslange I just ran `npm run build:rtl` before packaging for Beta 3

Issue #785, opened 2019-10-07
Accessibility-ready review


carolinan:
Default themes need to follow the requirements for the [accessibility-ready tag]
(https://make.wordpress.org/themes/handbook/review/accessibility/required/).

**Important:** The accessibility-ready requirements are **simplified,** all code added to core must conform with the [WCAG 2.0 guidelines at level AA. ](https://www.w3.org/TR/WCAG20/)

https://make.wordpress.org/core/handbook/best-practices/coding-standards/accessibility-coding-standards/

**Note 2**: this will most likely be updated to WCAG 2.1 in the near future so we should have this in mind.
https://www.w3.org/TR/WCAG21/

https://make.wordpress.org/accessibility/handbook/test-for-web-accessibility/

Reviews to complete for the **tag**:

- [ ] Keyboard Navigation
- [x] Controls
- [x] Skip Links
- [x] Forms
- [x] Headings
- [x] ARIA Landmark Roles
- [x] Content Links
- [ ] Repetitive Link Text
- [x] Contrasts
- [x] Images
- [x] Media
- [x] Screen Reader Text
- [x] Not Allowed

ntwb:
If possible it would be great to have Twenty Twenty WCAG 2.1 compatible

erikkroes:
The theme uses several HTML entities like `&larr;` throughout. For example, in pagination. They can be read by screen readers, but are functionally comparable to decorative images. 
Adding `aria-hidden` on the containing element, or adding a `<span>` with `aria-hidden` makes screen readers ignore them.
Fixed in #865

carolinan:
@pattonwebz The starter content does not follow the repetitive link requirement https://make.wordpress.org/themes/handbook/review/accessibility/required/#repetitive-link-text 
for the read more links:

`'<div class="wp-block-button is-style-outline"><a class="wp-block-button__link" href="https://make.wordpress.org/core/2019/09/27/block-editor-theme-related-updates-in-wordpress-5-3/">' . __( 'Read More', 'twentytwenty' ) . '</a></div>',`

joyously:
Is that a core Read More block? Or a button block that doesn't need to be there at all?

carolinan:
It is a button block that needs to be there.

carolinan:
Hi
I held it off because I am not 100% happy with the submenu toggles for the desktop horizontal menu. I still feel that the toggles should be separate from the parent menu item.
I have been unable to work on it.
[...]

PR #805, opened 2019-10-08
Accessibility improvements

erikkroes:
I've made 3 small changes that I think improve accessibility.

Html entities like `&larr;` are used in several spots. These can be confusing to screen reader users so I've made them aria-hidden.
I've changed the display of categories and tags to unordered lists, which should be a more fitting semantic.

It's my first contribution here, so I can imagine I missed something in code style.


erikkroes:
> Hi
> Can this be created as _issues first_ and split to separate PR's?`
> These points need to be reviewed and checked off here:
> #785

Will do when I can find the time, hopefully very soon!

PR #808, opened 2019-10-08
Makes menu text inherit color

glauberm:
Applying to `style-rtl.css` styles already applied to `style.css` as discussed [here](https://github.com/WordPress/twentytwenty/pull/723#discussion_r331746079).


ianbelanger79:
Already fixed in #809

PR #812, opened 2019-10-08
Conditional visibility for the cover template customizer section

### LINKED ISSUE #277: Conditional visibility for the cover template customizer section
The cover template settings in the customizer only apply to pages with the cover page template. For best user experience, the recommended practice is to only show settings when the page in the live preview can show the changes. Widgets use this logic in core. Without this functionality, the results of changing these settings are unclear to users.

The customize API includes the `active_callback` parameters in the control, section, and panel objects for this reason. We can add an `active_callback` to the `twentytwenty_cover_template_options` section registration, referencing a new function that detects whether the current page is using the cover page template.

acosmin:
Fix #277

@andersnoren 

A first try at this 😄 

- The Cover section will always remain active if it doesn't have any controls.
- A preview button will be added in the Cover section, when you click it it will preview the first post with a Cover template.
- The preview button appears only if the current preview doesn't have a Cover template.

When I have some time, I will do the part in case the website doesn't have any posts with the Cover template.


acosmin:
I could probably use the Customizer JS API to add the button + description, before I do that 😄 I need to know if this is something that is needs... so I don't waste time on it...

carolinan:
But _why_ ? Using an active callback is simpler. It doesn't create 115+ lines of JS...

acosmin:
This is based on @andersnoren comment: https://github.com/WordPress/twentytwenty/issues/277#issuecomment-539482574

There's no way for the Customizer to know that a post has a cover template, unless you send that information from the preview window.

carolinan:
is_page_template works for both posts and pages.

acosmin:
@carolinan I think I went down the rabbit hole 😄 

Thanks for bringing me back to reality 😆

Issue #791, opened 2019-10-07
The Read More link does not anchor to the Read More tag


ianbelanger79:
When using the Read More Block in a post, the Continue reading button does not link to the Read More tag within the post and it should.

joyously:
I can confirm that the fragment identifier is missing on the Read More links, both in WP 5.3 beta2 and WP 4.9.11. 
Using my own theme, both versions have the correct links, so it must be the theme.

PR #815, opened 2019-10-09
Fixed more tag output.

byalextran:
Fixes https://github.com/WordPress/twentytwenty/issues/791.

Is there a reason why the theme wouldn't use the default output from the more tag? What it's trying to do seems similar to what the default output does. 

This fix uses the default output and modifies it so the styling is correct.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -482,16 +482,14 @@ function twentytwenty_block_editor_settings() {
 add_action( 'after_setup_theme', 'twentytwenty_block_editor_settings' );
 
 /**
- * Read More Link
- * Overwrite default (more ...) tag
+ * Modify the more tag.
+ *
+ * @param string $link The default HTML output for the more tag.
+ *
+ * @return string $link The modified more tag.
  */
-function twentytwenty_read_more_tag() {
-	return sprintf(
-		'<a href="%1$s" class="more-link faux-button">%2$s <span class="screen-reader-text">"%3$s"</span></a>',
-		esc_url( get_permalink( get_the_ID() ) ),
-		__( 'Continue reading', 'twentytwenty' ),
-		esc_html( get_the_title( get_the_ID() ) )
-	);
+function twentytwenty_read_more_tag( $link ) {
+	return preg_replace( '/class="(.*)"/iU', 'class="$1 faux-button"', $link);
 }
 add_filter( 'the_content_more_link', 'twentytwenty_read_more_tag' );
 
--- a/functions.php
+++ b/functions.php
@@ -482,28 +482,6 @@ function twentytwenty_block_editor_settings() {
 add_action( 'after_setup_theme', 'twentytwenty_block_editor_settings' );
 
 /**
- * Modify the more tag.
- *
- * @param string $link The default HTML output for the more tag.
- *
- * @return string $link The modified more tag.
- */
-function twentytwenty_read_more_tag( $link ) {
-	$search = array(
-		'/class="(.*)"/iU',
-		'/<\/a>/iU',
-	);
-
-	$replace = array(
-		'class="$1 faux-button"',
-		sprintf( '<span class="screen-reader-text">"%1$s"</span></a>', esc_html( get_the_title( get_the_ID() ) ) ),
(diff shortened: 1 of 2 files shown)


byalextran:
on second look, the default output of `get_content()` with no parameters includes accessibility output. but if you do `get_content( 'Continue reading' )` no accessibility tags get added. i'm not sure why. 

regardless, i think adding the `Continue reading` text in the `get_content` function is the preferred way versus in the filter. so updated this PR to add the accessibility code as it was originally (not in the way `get_content()` does. 

```php
// default output from get_content()
<a href="http://localhost:8888/read-more-test/#more-50" class="more-link"><span aria-label="Continue reading Read More Test">Continue reading</span>
</a>

// twenty twenty output
<a href="http://localhost:8888/read-more-test/#more-50" class="more-link faux-button">Continue reading <span class="screen-reader-text">“Read More Test”</span></a>
```

joyously:
Why can't you just put the markup you want into the parameter passed to `the_content()`? Why use a filter?

byalextran:
@joyously looking at the supported parameters for `the_content()`, it looks like you can only change the text output (e.g. Read more, Continue reading, etc.) and not the markup itself. that's where the filter comes into play, to change the markup for accessibility purposes.

nielslange:
> @joyously looking at the supported parameters for `the_content()`, it looks like you can only change the text output (e.g. Read more, Continue reading, etc.) and not the markup itself. that's where the filter comes into play, to change the markup for accessibility purposes.

cc: @afercia

joyously:
>  it looks like you can only change the text output (e.g. Read more, Continue reading, etc.) and not the markup itself.

No, it looks like you pass in what you want and it is used as is. If you pass `null` then some accessible default is used. The filter is for code that isn't calling the function directly, but the theme _is_ calling it directly.

byalextran:
@joyously i think i'm following what you're saying. the original code used `the_content()` without a parameter and relied on the filter to output the `Continue reading` text. you're saying we should remove `Continue reading` from the filter and use it as a parameter to `the_content()`? 

if that's the case, i'm in agreement and is what i've done with this PR. 

and it does look like more parameter for `the_content()` will pass the text along as is. so we could get rid of the filter entirely with something like this: 

```php
the_content( sprintf( '<span class="faux-button">%1$s</span> <span class="screen-reader-text">"%2$s"</span>', __( 'Continue reading' ), get_the_title() ) );
```

if that's the best solution here, i can update the PR. just want to make sure it's the right approach for the theme and accessibility.

joyously:
For the text and its markup, yes, the parameter is better than the filter. 
As for the `faux-button` class, you'll have to test whether it should be on the link or on the text, but I don't see why the text would be a problem.

byalextran:
alright, i've updated this PR to use just `the_content()` without the need for the more tag filter. based on my quick test, using `.faux-button` on the `span` seems to work. 

it actually mirrors the hover state of the button block (text gets underlined) whereas with the class on the anchor there was no change on the hover state. 

however, would love someone more familiar with the styling/codebase confirm there won't be any issues with putting the class on the `span` nested within the anchor as opposed to the `a` itself.

carolinan:
Perhaps the filter was used because `the_conten`t is used in more than one place. This only addresses it in one place.

byalextran:
@carolinan a quick search of `the_content` in the codebase shows it in two spots. one in `content.php` and the other in `content-cover.php`. the former's description says "Used for both singular and index." the latter doesn't say anything which leads me to believe it wouldn't be used in an index context (and therefore wouldn't need to display the more tag). 

can anybody think of a use case for the cover template being used as an index page?

byalextran:
i just did a little playing around in the homepage customizer settings. i created a page using the cover template and set it as the posts page. when viewing that page, the index view is from `content.php` and not `content-cover.php`.

so it appears using just `the_content()` (without the more tag parameter) in `content-cover.php` is fine. but again, i defer to someone with a deeper understanding of downstream impact if we also need to add the more tag parameter there or revert back to using a filter.

PR #819, opened 2019-10-09
Fix escaping function

le-prince-noir:
It is possible to add scripts when translating strings.
Adding an escaping function corrects the problem.

Cheers,
Samy


carolinan:
Hi
We have had several discussions regarding escaping translations and we will not be escaping them when they are between HTML tags.

https://github.com/WordPress/twentytwenty/issues/176

PR #825, opened 2019-10-10
Declare all starter content

Clorith:
WordPress core does not (at this time) provide block-aware starter content by default, so declaring them all manually is needed to avoid page edits starting with the classic block.

A core ticket exists at https://core.trac.wordpress.org/ticket/48130 to update the starter content for the new age of blocks, but to ensure the best possible user experience it would be nice to provide blocks with Twenty Twenty regardless of when the core patch lands.

This PR incorporates the starter content from the patch planned for core, the only difference is that it has the `twentytwenty` text domain added here.


joyously:
The change has been committed to ticket 48130 now: https://core.trac.wordpress.org/changeset/46457

nielslange:
> Quick update that the core ticket has made it in, but for those who may not be using WordPress 5.3, it might still be useful to have this added directly by the theme for completeness sake 🤔

Would providing two started content sets, one for `>= 5.3` and one for `<5.3` make sense?

joyously:
I think the theme should just use the core content, to be generic, and not worry about whether that is blocks or not. The theme should be block-agnostic.

carolinan:
This also adds more text after the string freeze.

Clorith:
@nielslange I'd say it doesn't make sense, since the block-declarations would be hidden and not have any effect in the classic editor, but the text would still be available.

I'm running Twenty Twenty on a 5.2 site right now for example, one that's unlikely to get updated to 5.3 on release, but the theme worked really well for that sites use case, that's the scenario I'd love for this to cover.

Clorith:
@carolinan string freeze isn't until RC1, so we still have a couple of days :)

carolinan:
> @carolinan string freeze isn't until RC1, so we still have a couple of days :)

Soft string freeze in place since october 8. I have not seen any other announcements?
https://make.wordpress.org/core/5-3/
https://make.wordpress.org/polyglots/2019/10/03/wordpress-5-3-ready-to-be-translated/

joyously:
@Clorith but starter content only applies to fresh sites.

garrett-eclipse:
Can we not just drop the starter content from the theme and have it use core. With 5.3 having committed the block-enabling change in [Changeset#46457](https://core.trac.wordpress.org/changeset/46457) new users will already get block-enabled starter content on install.

I understand @Clorith we wanted to be more encompassing with non-5.3 users who decide to install TwentyTwenty, but in most of those cases these users already have content on site making it it 'dirty' (setting `fresh_site` to 0) which disables starter content. I would be surprised if someone is starting a new site with an old version of WP specifically to use the new version of the default theme. And for those very odd edge-cases they would just get a classic block, nothing would be broken.

Clorith:
I tend to agree with you there @garrett-eclipse, very good point. Let's shelve this for now, if we find it to be an issue we can instead revisit at such a time.

PR #826, opened 2019-10-10
Make regular the default font size

### LINKED ISSUE #821: Block Editor Font Size Issues: Medium option missing, wrong value for Default and Large options
**1)** In style.css there is a `.has-medium-font-size` class but there is no Medium option 
in the block editor. https://github.com/WordPress/twentytwenty/blob/f55d056701f3bef1fc0d4c9245983ea30e90cac2/style.css#L2675-L2698

block-editor

---

**2)** In style.css the `.has-large-font-size` is set to 1.25em (= 26.25px) while inside the block editor the indicated value is 24.

24px

---

**3)**  Inside the block editor, for the paragraph block the default value is set to Small instead of Regular.

carolinan:
Make regular the default font size.
Increase size of the Large font size.
Attempts to fix #821 

(CSS for medium is left as backup for previously created blocks).


functions.php
```diff
@@ -444,22 +444,22 @@ function twentytwenty_block_editor_settings() {
...
+				'name'      => _x( 'Small', 'Name of the small font size in the block editor', 'twentytwenty' ),
+				'shortName' => _x( 'S', 'Short name of the small font size in the block editor.', 'twentytwenty' ),
+				'size'      => 18,
+				'slug'      => 'small',
+			),
 			array(
 				'name'      => _x( 'Large', 'Name of the large font size in the block editor', 'twentytwenty' ),
 				'shortName' => _x( 'L', 'Short name of the large font size in the block editor.', 'twentytwenty' ),
-				'size'      => 24,
+				'size'      => 26.25,
```
nielslange: Do we need to make the large font size even larger?

yannickiki:
Current PR | Gutenberg Default
-- | --
current | default

The current PR changes the ascending order of the font size dropdown (Regular should be after Small).
A solution is to change the `'regular'` slug to `'normal'`

@nielslange I think the size of the large font is good at 26,25, as the regular is set to 21 and the larger one to 32.

carolinan:
But how would changing the name from Regular to Normal affect the string freeze?
-Because it is also weird to have the slug _normal_ and the name _Regular_.
Should Larger be changed to Huge?

functions.php
pattonwebz: This new size makes the size in editor view match the size on the frontend. Frontend calculates to 26.25px

ianbelanger79:
I updated the PR with `Small` being the first option in the dropdown. I think this is the least confusing option.

yannickiki:
@ianbelanger79 The commit set `Small` as the default font-size instead of `Regular`. It's confusing because every time the user inserts a new paragraph block, it shows that the font size is set to `Small`  which is not true (it's actually `Regular`).

PR #827, opened 2019-10-10
replace .var with -var for variable fonts

### LINKED ISSUE #718: Getting font related error in console while in customizer
I am getting getting the following font related error in console while in customizer.
image

carolinan:
Replaces .var in the name of the variable font -var.
Fixes #718


pattonwebz:
This does resolve the issue I was seeing in the console and allows the fonts to show up :)

This PR adds new files with the new name and does not remove the older files. Not sure why git didn't track it as a rename but maybe cos it's a binary file?

joyously:
Git doesn't track renames. Only delete and add. (You need something like Bazaar to track renames.)

PR #831, opened 2019-10-10
Remove unnecessary HTML output 

### LINKED ISSUE #798: Remove unnecessary HTML output for the desktop header toggles (search & menu buttons) if they aren't set
Insider `header.php`, the `.header-toggles` div shouldn't be output if search is deactivated in the customizer and if there's no "Desktop Expanded Menu" set.
Also if the search is activated but there's no "Desktop Expanded Menu" set, the` .nav-toggle-wrapper` shouldn't be output.

carolinan:
Hopefully fixes #798 I hope my logic did not fail me.


dianeco:
I've tested, looks good to me.

Issue #833, opened 2019-10-11
After closing modals, aria-expanded is still set to true


carolinan:
When closing the search modal, the aria-expanded of the _search-toggle mobile-search-toggle_
remains set to true, it needs to be set back to false.

When closing the the expanded menu or the mobile menu (both assigned and fallback)
the aria-expanded of the _nav-toggle mobile-nav-toggle_ remains set to true, it needs to be set back to false.

carolinan:
If I understand this correctly, it is because the toggles are only pressed once, they are not used to close the modals.  Similarly, the close buttons will only be pressed once. 
So using the JS toggle attribute will not work here. 
https://github.com/WordPress/twentytwenty/blob/master/assets/js/index.js#L646
https://github.com/WordPress/twentytwenty/blob/master/assets/js/index.js#L771

ianbelanger79:
Looks like this was fixed by #835

PR #835, opened 2019-10-11
Fix - After closing modals, aria-expanded is still set to true

acosmin:
Fix for #833

This needs to be done regardless of what happens in #834


carolinan:
Ok so it seems to be working correctly for the submenu items in the expanded menu. But not for the modals toggle buttons.

acosmin:
@carolinan hope this one does it 😄 it's kinda hard to debug when you didn't write the initial code...

acosmin:
I don't think so, it looks the same :)

PR #837, opened 2019-10-11
Make the CSS for hiding the RSS icon stricter

### LINKED ISSUE #836: RSS Widget has all links hidden
The styling at https://github.com/WordPress/twentytwenty/blob/master/style.css#L4021-L4023 makes all links in the RSS feed widget hidden.

The intention was to remove the RSS icon (see #709), but it has the side-effect of hiding every feed link as well.

Clorith:
This makes sure the style rule added in #709 only affects the title section for targeting the RSS feed icon.

Fixed #836

### DIFF
--- a/style.css
+++ b/style.css
@@ -4018,7 +4018,7 @@ div.comment:first-of-type {
 
 /* Widget: RSS ------------------------------- */
 
-.widget_rss a.rsswidget:first-of-type {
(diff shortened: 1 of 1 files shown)


andersnoren:
@Clorith Thanks, merged! What's your WordPress.org username?

Clorith:
@andersnoren That would be `Clorith` (capital `C`)

Issue #832, opened 2019-10-11
The mobile menu does not constrain focus when a fallback is used.


carolinan:
**When no menu is assigned:**
**Desktop:** A one level fall back menu is used. No issues found.
**Mobile:**
The menu can be opened with keyboard.
Items can be navigated through with keyboard.
However, the focus is not constrained within the menu modal. When you tab forwards from the last menu item, an item on the page below the menu is targeted. You can't see the focus because the page is covered by the open menu.
-It is possible to tab backwards to close the menu using the button, but it is not intuitive.

acosmin:
@c4rl0sbr4v0 yes, please do 👍

andersnoren:
It looks like none of the modal menus (mobile or desktop, fallback or not fallback) constrain focus currently. Could [this conditional](https://github.com/WordPress/twentytwenty/blob/master/assets/js/index.js#L444) in `goBackToCloseButton` be causing it?

```
mobileMenu = document.querySelector( '.mobile-menu' );

if ( mobileMenu ) {
	return false;
}
```

There's always an element with the `.mobile-menu` class output, so I think that conditional always returns false.

acosmin:
@andersnoren wow :) that's a huge type, sorry for that, I meant for it to be `if( ! mobile...`

andersnoren:
@acosmin No worries, it happens 🙂

PR #845, opened 2019-10-12
Fix - The mobile menu does not constrain focus when a fallback is used

acosmin:
Fix for #832


carolinan:
The focus does not stay within the search modal when it is open.
This means that when you tab backwards you end up below the modal and you can't see where you are.

acosmin:
@carolinan give it a try now :)

acosmin:
I think the script can be simplified, I'll take a look

acosmin:
@carolinan Refactored 💯 😃

acosmin:
Finally 😄  adding so many menu locations is a pain from a code point of view, I'm starting to regret voting for this solution 😆

assets/js/index.js
```diff
@@ -435,74 +435,29 @@ twentytwenty.modalMenu = {
...
-			lastModalMenuItems = hasSocialMenu ? socialLinks : menuLinks;
-			focusedElementParentLi = twentytwentyFindParents( event.target, 'li' );
-			focusedElementIsInsideModal = twentytwentyFindParents( event.target, '.menu-modal' ).length > 0;
+				elements = modal.querySelectorAll( 'input, a, button' );
+				elements = Array.prototype.slice.call( elements );
 
-			lastMenuItem = lastModalMenuItems[lastModalMenuItems.length - 1];
+				lastEl = elements[ elements.length - 1 ];
+				firstEl = elements[0];
+				activeEl = document.activeElement;
```
c4rl0sbr4v0: Are these variables taking into account that the last menu item can have submenus open or closed?

acosmin: I hope so 😆

acosmin: You're right about that, I'll have to fix it 😄

acosmin: Fixed in 6a9fec2

Please check it if you have the time 😄

c4rl0sbr4v0: You are making a huuuge refactor 😆 , let you know if fails.

c4rl0sbr4v0: Works like a charm :-)

acosmin: thanks 😄 marking this as resolved

carolinan:
I have tested in Chrome, Firefox, IE11 and Edge, also on 4.9 and on android mobile and Ipad (Chrome, Safari). I have limited experience with using screen readers but I tested with NVDA and it seems to work well.

PR #849, opened 2019-10-14
Set the menu modal overflow-x to hidden

briceduclos:
There is a bug on Firefox when opening the modal menu. It displays a horizontal scrollbar.
scrollbar

This PR only changes the modal menu overflow to fix this bug.


ianbelanger79:
Thanks for the PR @briceduclos

What is your WP.org username? So we can give you props.

briceduclos:
@ianbelanger79 My WP.org username is [bduclos](https://profiles.wordpress.org/bduclos/)

Issue #289, opened 2019-09-15
Add an option inside the customizer to display the title on the main posts page


yannickiki:
Currenty the main posts page doesn't display the page title. It would be a nice feature to have the possibility to display the page title like "Blog", "Ideas & thoughts", "Articles"... and would add consistency as all the other pages display the title.

joyously:
My vote is that it's not needed.

nielslange:
Thanks for suggesting this, @yannickiki. I kind of like the idea to show the title on the blog posts page, but would not implement it as part of https://github.com/WordPress/twentytwenty/projects/1.

yannickiki:
@nielslange or maybe instead of an option,  adding a filter in the index.php file could do the trick:
`$archive_title = apply_filters( 'twentytwenty_get_the_archive_title', $archive_title );`

I think it would be a good compromise.

nielslange:
As mentioned before, I like the idea. But given the tuff timeline of the Beta 1 release, I would address this enhancement later.

celloexpressions:
I agree that this is not needed. A filter seems reasonable, but a user-facing option will have more consequences than benefits for usability.

dianeco:
@nielslange I agree with the comments above, a filter would be a nice & reasonable feature. I've created a test site where all the pages have a color banner (home, about, contact, single post, archive) except for the blog index which kind of breaks the consistency of the design. In this case a filter to display the title on the blog page would have been useful.

PR #855, opened 2019-10-15
Add filter for the archive title

dianeco:
As previously mentioned in #289, a filter for the archive title would be a nice and reasonable enhancement for developers. Themes like Chaplin or Twenty Seventeen display the title on the main posts page.

### DIFF
--- a/index.php
+++ b/index.php
@@ -51,6 +51,8 @@ get_header();
 		$archive_title    = get_the_archive_title();
 		$archive_subtitle = get_the_archive_description();
 	}
+	
+	$archive_title = apply_filters( 'twentytwenty_archive_title', $archive_title );
 
 	if ( $archive_title || $archive_subtitle ) {
 		?>


index.php
```diff
@@ -51,6 +51,8 @@
 		$archive_title    = get_the_archive_title();
 		$archive_subtitle = get_the_archive_description();
 	}
+	
+	$archive_title = apply_filters( 'twentytwenty_archive_title', $archive_title );
```
pattonwebz: This needs a code comment to provide relevant filter documentation.

pattonwebz:
Hey @dianeco I was just about to move this PR over to the main WP trac instance when I realised that we would not need this filter in the theme as `get_the_archive_title()` has a filter at the end which you could use instead to do this :)

- https://developer.wordpress.org/reference/functions/get_the_archive_title/
- https://developer.wordpress.org/reference/hooks/get_the_archive_title/

PR #857, opened 2019-10-16
Removed some unused variables

itowhid06:


pattonwebz:
Should we be thinking about why these variables are no longer used and checking if it is unintentional? PR itself looks good though :)

PR #865, opened 2019-10-17
Hide decorative HTML entities

erikkroes:
Fixes the use of decorative HTML entities as mentioned in #785


carolinan:
@ianbelanger79 this can't be a future consideration because the default theme has to be accessible.

PR #868, opened 2019-10-17
Prevent filtering edit_post_link in the admin

westonruter:
In the AMP plugin there are links to in the classic publish metabox for viewing and editing a given URL, looking like this:

> Screen Shot 2019-10-17 at 09 18 47

When Twenty Twenty is active, this appears broken:

> Screen Shot 2019-10-17 at 09 18 31

The `twentytwenty_edit_post_link` function should short-circuit when `is_admin()`. This PR implements that change.

Introduced in #449.


westonruter:
Perhaps there are other such filters which should short-circuit in an admin contect, but this is the one that jumped out to me.

PR #869, opened 2019-10-17
Remove custom starter content

### LINKED ISSUE #38: Add/improve starter content

garrett-eclipse:
As discussed in #825 and on [Slack](https://wordpress.slack.com/archives/C02RQBWTW/p1570726252294400) this removes the custom starter content to use the default core content which is more re-usable.
Fixes #38


ianbelanger79:
The team has decided to keep our custom starter content

garrett-eclipse:
Thanks @ianbelanger79

Issue #828, opened 2019-10-10
Desktop Expanded Menu vanishes on smaller screen replaced by Desktop Horizontal Menu


garretthyder:
Hello,

While testing starter content I noticed that on desktop we see 'Contact' visible form the Desktop Horizontal Menu and if we toggle the menu we see the 'Home', 'About' and 'Blog' from the Desktop Expanded Menu.

But when we reduce the screensize and the menu adapts for mobile there's no menu items visible and when you toggle the menu only 'Contact' from the Desktop Horizontal Menu is available and we've completely lost the Desktop Expanded Menu (Home/About/Blog).

![bc2e953900abf95176c57c362997161f](https://user-images.githubusercontent.com/8726005/66592695-6c624e80-eb4a-11e9-962f-65bda5c22700.gif)

On mobile could we preserve the Desktop Expanded Menu and simply append the Desktop Horizontal Menu to the bottom of it?

Thanks

carolinan:
Would that not defeat the purpose of having a mobile menu?

garrett-eclipse:
Thanks @carolinan if the goal is to have a mobile specific menu shouldn't there be a dedicated menu location for it?

Currently, it seems odd why we wouldn't just use the expanded menu for the mobile menu so the contents of that menu are consistent across all devices. Swapping the menu contents for the horizontal menu also seems limiting as you'd most likely only place a few important links there and have your full menu in the expanded menu.

joyously:
You can read through the menu discussion on Slack, back in September, where we hashed all this out.
The mobile menu does have its own menu location. But it uses the main menu as a fallback if it's empty.
I'm not sure I understand it well enough to explain further.

garrett-eclipse:
Thanks @joyously using the Mobile menu location on a new menu works nicely and clears things up. SO I guess my qualm is more with when nothing is using the mobile location it currently defaults to use the menu set to the Desktop Horizontal Menu when I feel it would make more sense to default to the Desktop Expanded Menu when the Mobile menu location isn't set.

andersnoren:
Hi @garrett-eclipse,

Since the horizontal menu is the more traditional one, the decision was made to make that one the default menu location in the theme. If a user switches to Twenty Twenty from a different theme, any menu set to one of the following menu locations in their previous theme will automatically be moved to the `primary` menu location (Desktop Horizontal Menu) in Twenty Twenty:

```
[...]

PR #876, opened 2019-10-18
Move starter_content to primary menu to avoid vanishing on mobile

garrett-eclipse:
Update #828 to address the starter_content vanishing on mobile with the default menu setup. This will remove the expanded desktop menu and place all items in the primary menu. With the current starter_content setup most items are added to the 'Desktop Expanded Menu' which vanishes on mobile.


ianbelanger79:
Closing this PR as it is failing travis and we cannot change the starter content at this point.

garrett-eclipse:
Thanks @ianbelanger79 shall I refresh for 5.3.1?
It's unfortunate we'll be unable to address this critical issue before the theme is released. As it stands any user adopting the existing starter content will have the majority of their content inaccessible on mobile.

pattonwebz:
Reopening this so it can be worked on some more for future release. It's been flagged a few times that the current menu setup is confusing for people so we should do something about this soon (and maybe we can squeeze this one in as a need for UX reasons but if not it can go after release_

garrett-eclipse:
I replaced this corrupt branch with #955

PR #337, opened 2019-09-17
TRY #118: Support for non-latin font fallbacks

nielslange:
Fixes #118 

⚠️ Thir PR is still in progress and I only pushed the changes so that the direction of this PR can be discussed.

PR #891, opened 2019-10-19
Updated support for non-latin fonts

andersnoren:
Fork of #337. Updates that PR by:

- Removing the font files
- Fixing the merge conflicts with master
- Updating the Japanese font family stack, per comments in #118
- Adding enqueue of block styles and editor styles, in addition to the front-end ones
- Updating the structure to be somewhat closer to `twentytwenty_get_customizer_css`

Tested with a number of different languages on the front-end and in the block editor and classic editor.


classes/class-twentytwenty-non-latin-languages.php
```diff
@@ -0,0 +1,125 @@
...
+			// Fetch users locale.
+			$locale = get_bloginfo( 'language' );
+
+			// Define fallback fonts for non-latin languages.
+			$font_family = apply_filters(
+				'twentytwenty_get_localized_font_family_types',
+				array(
+
+					// Arabic.
+					'ar'    => array( 'Tahoma', 'Arial', 'sans_serif' ),
```
carolinan: Is this meant to be sans-serif? Not sans_serif?

andersnoren: Those should all be `sans-serif`. Updated.

PR #895, opened 2019-10-19
Fix block button text color setting not taking effect

### LINKED ISSUE #893: Button block: Text color setting with editor palette colors doesn't work
The text color setting for the button block currently doesn't work in Twenty Twenty when using the colors in the editor palette. This is caused by the `.wp-block-button__link` color styling not accounting for the `.has-text-color` class, indicating a text color overwrite, and for the default block styles setting `.wp-block-button__link:active, .wp-block-button__link:focus, .wp-block-button__link:hover, .wp-block-button__link:visited` to `color: inherit`, which needs to be overwritten with styling with higher specificity.

Buttons with custom text colors (using hex values) work.

The images show a button with the background color set to "Subtle Background" and the color set to "Accent Color".

Currently in the editor | Currently on the front-end
------------ | -------------
image | image

andersnoren:
As described in #893, the button block text color setting isn't working with the editor palette classes. This PR fixes that issue. 

The root of the problem is that the `.has-x-color` styling is being overwritten by `color: inherit` being applied to the button element with higher specificity. In the theme, we can fix this by updating the targeting to `:not(.has-text-color)`, which this PR does. 

However, the core block editor styles (output on the front-end) includes the following CSS:

```
.wp-block-button__link:active, .wp-block-button__link:focus, .wp-block-button__link:hover, .wp-block-button__link:visited {
    color: inherit;
}
```

Which also overwrites the `.has-x-color` class, applied to the `.wp-block-button__link` element. I couldn't see any solution to this other than reapplying all of the `.has-x-color` classes to this element specifically, with a specificity high enough that it overwrites the `color: inherit` block above. So this PR adds the following:

```
.wp-block-button__link.has-accent-color {
	color: #cd2653;
}

.wp-block-button__link.has-primary-color {
	color: #000;
}

.wp-block-button__link.has-secondary-color {
	color: #6d6d6d;
}

.wp-block-button__link.has-subtle-background-color {
	color: #dcd7ca;
}

.wp-block-button__link.has-background-color {
	color: #f5efe0;
}
```

Along with corresponding updates to the targeting in `twentytwenty_get_elements_array`. This is an **ugly** solution, but I don't see another way of overruling the `color: inherit` set in the core styles.

---

**Edit:** Updated the solution per [this suggestion](https://github.com/WordPress/twentytwenty/pull/895#issuecomment-544211754) by @m-e-h, which brings it more in line with how core handles the specificity of the color classes.

Fixes #893.


joyously:
Another solution is to get the editor styles changed. 
I've been attempting that for quite awhile now (see https://github.com/WordPress/gutenberg/issues/12299), and so has @m-e-h. (See https://github.com/WordPress/gutenberg/issues/10178 and https://github.com/WordPress/gutenberg/issues/11779)

m-e-h:
Maybe just strengthen the utility color classes in general. That's how core ended up dealing with it. 
For example, they use `:root .has-luminous-vivid-amber-color {` which overrides the `inherit` style.

Discussion on that decision and some other options are here:
https://github.com/WordPress/gutenberg/issues/12986#issuecomment-485890449

andersnoren:
@m-e-h Thanks for the link, and the suggestion! Increasing the specificity of the color classes by prefixing them with `:root` solved the problem, and allowed us to remove some extra targets from `twentytwenty_get_elements_array` to boot.

PR #928, opened 2019-10-26
Fix scroll when clicking "Reply" on a comment

### LINKED ISSUE #927: comment reply link doesn't work right
(Mentioned in today's accessibility bug scrub)
On a post with several comments, click the Reply links.

**Expected result:** either the page is scrolled to the comment form at the bottom, or the form appears at the comment where Reply was clicked.

**Actual result:** the page is scrolled to a random-seeming location and the comment form may or may not be visible.

andersnoren:
Fixes the `classList.contains()` conditional in `twentytwenty.smoothScroll.scrollToAnchor`, which was causing the smooth scroll to trigger when clicking an element with the `.do-not-scroll` class. Also updates the conditional with a check for `#cancel-comment-reply-link`, making sure that the smooth scroll doesn't trigger when the comment respond view is closed.

Fixes #927.


andersnoren:
Removed changes to `style.css`.

PR #930, opened 2019-10-26
Fix author box not matching design

### LINKED ISSUE #929: Post author box doesn't match design
PR #874 caused the layout of the author box to change from the design, resulting in the text in the author box to have less horisontal space on mobile and the title (author name) missing a bottom margin. I'll create a PR for this during the day that reverts the changes in #874.

andersnoren:
Reverts #874 changing the design of the author box, while also fixing the HTML warnings in the original code. Fixes #929.

Before | After
------------ | -------------
image | image


andersnoren:
Good catch, updated. I missed that the `aria-hidden` span had been added to master.

Yeah, it matches the capitalization used elsewhere in the theme ("Older Posts" etc).

PR #931, opened 2019-10-27
Update table styles

### LINKED ISSUE #320: table style : editor versus front end

carolinan:
Fixes #320
Removes table-layout: fixed; unless the table is fixed width.
Updated font size, padding and table header default text alignment


joyously:
Are you aware that the table block markup changed for 5.3? 
Was this style tested in other versions?
And I don't see a change for the border color.

carolinan:
I have only tested in 5.3 RC2.
The border is #dcd7ca both on the front and editor.

PR #941, opened 2019-10-31
Fixed bottom border on striped tables

byalextran:
Noticed the bottom border is removed on striped tables. 

Screen Shot 2019-10-30 at 21 02 47 PM

PR fixes that. 

Screen Shot 2019-10-30 at 21 03 06 PM

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -72,3 +72,6 @@ Version 1.0:
 | @josephscott | @josephscott |
 | @Swapnildhanrale | @swapnild |
 | @Aurorum | @torres126 |
+| @erikkroes | @erikkroes |
+| @mehidi258 | @mehidi258 |
+| @briceduclos | 
--- a/README.md
+++ b/README.md
@@ -73,6 +73,9 @@ Source: https://wordpress.org/themes/twentyseventeen/
 Included as part of the following classes and functions:  
 - twentytwenty_unique_id()  
 
+Underscores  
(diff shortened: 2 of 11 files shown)


byalextran:
i went ahead and created this PR since it seemed like an easy fix. do i need to submit a reference issue?

pattonwebz:
@byalextran thanks for submitting this, we do not NEED an issue for things (but sometimes that helps track it). 

This CSS change here does fix the border missing at the end of striped tables but I am not sure if there are other implications or the reason that the CSS existed in the first place.

andersnoren:
@byalextran It seems like this issue stems from the `.wp-block-table.is-style-stripes` classes being applied to the `table` element in 5.2, and to a wrapping `figure` element in 5.3. 

On 5.2 (which it looks like you're running), we don't want to remove the bottom border from the `.wp-block-table.is-style-stripes` element, since the classes are on the `table` element. On 5.3, with the classes applied to the wrapping `figure`, we do want to remove the bottom border (since not doing that will create an extra border beneath the table element).

If you restore the CSS you removed in this PR, and change its targeting to only hit the figure element, we should be good.

```
figure.wp-block-table.is-style-stripes {
    border-bottom: none;
}
```

The change in markup structure in 5.3 also causes the striped style to lose the border around the table itself, as shown in the screenshot below:

image

If you add the following CSS, we should get our table border back on the striped style:

```
.wp-block-table.is-style-stripes table {
    border-collapse: inherit;
}
```

image

byalextran:
@andersnoren great, thanks for the direction on the proper fix! just updated the PR.

PR #944, opened 2019-11-01
Remove margins for the group block when a background-color is set

### LINKED ISSUE #943: Impossible to display 2 consecutive full-width group without margins (contrary to Core blocks and Twenty Nineteen)
A common layout that we see on websites is a layout with 2 or more consecutive full-width colored group.

consecutive-groups

With Twenty Twenty, this layout is impossible to achieve as margins of 10 rem are applied on elements with the alignfull class.

The Core blocks default styles as well as Twenty Nineteen set the top and bottom margins to 0 when a background-color is set. 

https://github.com/WordPress/WordPress/blob/979a52690290f2134e2ab6129f5c443112c9df2b/wp-includes/css/dist/block-library/theme.css#L115-L118

https://github.com/WordPress/WordPress/blob/979a52690290f2134e2ab6129f5c443112c9df2b/wp-content/themes/twentynineteen/style.css#L6140-L6144

Moreover, if the user wants to add spacing between the groups, the Spacer block can be used to achieve it.

briceduclos:
This PR fixes #943

### DIFF
--- a/style.css
+++ b/style.css
@@ -2992,6 +2992,8 @@ figure.wp-block-gallery.alignfull {
 
 .wp-block-group.has-background {
 	padding: 2rem;
+	margin-top: 0;
+	margin-bottom: 0;
 }
 
 .wp-block-group__inner-container {


pattonwebz:
@andersnoren I think you would need to look at this one. I am unable to decide on my own if this is the best way to handle this or if there are more things that may need considered with this kind of change.

I am going to mark this for moving to trac because I do not think this will be able to be checked in time for release. We can look at it after WP 5.3 releases again but that time it will be on trac so I will move this ticket over and make a patch from the PR.

PR #946, opened 2019-11-01
Ensure link to read more is preserved in Continue reading button

### LINKED ISSUE #945: "Read More" tag does nothing on click
https://github.com/WordPress/twentytwenty/blob/41b4bf3f5969d2c581099240b8b44e2536226b44/functions.php#L568

This swallows the attributes of the link and outputs it as a span, which does not link to anything.

A solution would be to reinstate a call to `get_permalink( get_the_ID() )` as per the deleted code in [this PR](https://github.com/WordPress/twentytwenty/pull/815) that introduced the bug, or simply not to swallow the anchor tag in its entirety.

https://github.com/WordPress/twentytwenty/pull/815

gravityrail:
Fixes #945 

The filter which modifies the "Read More" link ("Continue reading" in twentytwenty) was stripping out the URL for the extra content on the post page.

This PR preserves the link and modifies styles so that link appears correctly (default anchor styles were making the text red instead of white)

To test:

* On any feed page, click the "Continue reading" button. You should be taken to the post.
* On master, the continue reading button is not a link, so it does nothing.

### DIFF
--- a/style-rtl.css
+++ b/style-rtl.css
@@ -696,11 +696,16 @@ button,
 	button,
 	.button,
 	.faux-button,
-	.faux-button.more-link,
+	.faux-button .more-link,
 	.wp-block-button__link,
 	.wp-block-file__button {
 		font-family: "Inter var", -apple-system, BlinkMacSystemFont, "Helvetica Neue", Helvetica, sans-serif;
 	}
+}
+
+.faux-button .more-link {
+	color: #fff;
+	text-decoration: none;
 }
 
 input,
@@ -811,6 +816,8 @@ button:hover,
 .button:hover,
 .faux-button:focus,
 .faux-button:hover,
+.faux-button .more-link:focus,
+.faux-button .more-link:hover,
 .wp-block-button .wp-block-button__link:focus,
 .wp-block-button .wp-block-button__link:hover,
 .wp-block-file .wp-block-file__button:focus,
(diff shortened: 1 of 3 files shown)


style-rtl.css
```diff
@@ -811,6 +816,8 @@ button:hover,
 .button:hover,
 .faux-button:focus,
 .faux-button:hover,
+.faux-button .more-link:focus,
```
desrosj: Is there a scenario where this is required? Without this addition, the read more buttons seemed to work perfectly fine.

gravityrail: Did you test this line with an RTL locale?

I find that without this line, the `.faux-button .more-link` removes the underline and then it never gets added back.

byalextran:
i believe the extra CSS is unnecessary with this approach: 

```php
return preg_replace( '/<a(.*)>(.*)<\/a>/iU', sprintf( '<a$1><span class="faux-button">$2</span> <span class="screen-reader-text">"%1$s"</span></a>', get_the_title( get_the_ID() ) ), $html );
```

pattonwebz:
I am adding the `future consideration` tag to this because the changes here include a CSS adjustment that we did not port into the mini-fix we rolled in today. While we fixed the broken link in that one there are still things to consider from this PR that we may want to add after release.

PR #952, opened 2019-11-05
Update Ensure link to read more is preserved in Continue reading button 

carolinan:
credits gravityrail, byalextran

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -566,7 +566,7 @@ add_action( 'after_setup_theme', 'twentytwenty_block_editor_settings' );
  * @return string $html
  */
 function twentytwenty_read_more_tag( $html ) {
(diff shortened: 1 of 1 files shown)


carolinan:
Underline works well on RTL, and the link works.
[

](url)

joyously:
> Tested with custom colors, RTL, custom read more text.

Is it the Read More or is it the Continue Reading? Be sure the right thing was tested.

carolinan:
I have tested the read more block.

andersnoren:
Same.

carolinan:
I can't get the other continue reading link to show actually. It only shows me ` […]`

joyously:
It seems to me that the code that was changed is the filter for the parameter passed to `the_content()`, which is *not* the Read More tag (or block), and is a strange way for the theme to do it since there is no option for the user to change the text. The theme would simply pass a different argument to `the_content()` instead of using a filter, unless it is intent on using the core default argument.

PR #953, opened 2019-11-05
Adjust regex to ensure links are not stripped from 'Continue Reading'

pattonwebz:
This PR is a mini fix for the PR open in #946. The method is taken from https://github.com/WordPress/twentytwenty/pull/946#issuecomment-548964756

Credits for this fix go to @byalextran and @gravityrail and the other PR may still be considered in future. This is just the absolute min required to correct the base problem but there may be more nuances to deal with (like styles).


pattonwebz:
Looks like I was too slow and this already got solved :)

Issue #961, opened 2019-11-06
Full Width Columns are missing horizontal spacing


yannickiki:
If the user inserts a full width Columns block, there's no spacing between the text and the edges of the screen. For devices that have a scrollbar, the edges of the content are outside the screen.

![fullwidthcolumns](https://user-images.githubusercontent.com/54663751/68326355-90726a80-00cb-11ea-96ff-e32371e65a53.png)

carolinan:
This was supposedly fixed with #707. To reproduce, did you just change the width of the starter content, or?

andersnoren:
I'll include a fix for this as part of the fix for #963. Currently, only paragraphs get the horizontal padding on the first and last items, which results in the spacing being uneven (described in #963) and the full-width columns block as a whole missing a horizontal padding.

PR #967, opened 2019-11-06
Fixes for the Columns block

### LINKED ISSUE #960: Spacing issue when columns block is followed by another columns block
On medium devices (782-999px), there are no more spacing between the 2 columns blocks. This issue can be seen on http://2020.wordpress.net/ .

columns

### LINKED ISSUE #963: Paragraphs inside columns which are nested inside full width block have unwanted side padding
To reproduce this issue:
- insert a full width Group block (or a Cover block)
- inside the group, insert a Columns block
- inside a column, add a Paragraph block 

The paragraph has an unwanted side padding making it misaligned with the rest of the content. 
Common layouts like the one below are therefore broken:

paragraph

andersnoren:
This PR contains fixes for the Columns block issues created by @yannickiki earlier today.

Fixes #963 and #961 by removing the paragraph-only padding used for the first and last columns in full-width columns blocks previously, and replaces it with a horizontal padding on the block as a whole. This made it possible to reduce the amount of CSS quite a bit, since it removed the need for different solutions for different breakpoints.

Before | After
------------ | -------------
image | image

Fixes #960 by adjusting the spacing between stacked Columns blocks between `782px` and `999px`. Also tweaked the margin on desktop, to make the horizontal margin between stacked columns blocks `32px` (same as the gutter) on all screen sizes.

Before | After
------------ | -------------
image | image

### DIFF
--- a/style-rtl.css
+++ b/style-rtl.css
@@ -5277,6 +5277,11 @@ a.to-the-top > * {
 		margin-bottom: 0;
 	}
 
+	.wp-block-columns.alignwide + .wp-block-columns.alignwide,
+	.wp-block-columns.alignfull + .wp-block-columns.alignfull {
+		margin-top: -2.8rem;
+	}
+
 	.alignfull .wp-block-column:first-child > p,
 	.alignfull:not(.has-background) .wp-block-group__inner-container > p:not(.has-background-color) {
 		padding-right: 2rem;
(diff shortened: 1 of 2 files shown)


carolinan:
@yannickiki Do you have time to test this? :)

PR #949, opened 2019-11-04
touch devices: open submenus on the desktop horizontal menu

carolinan:
Hopefully Fixes #790

PR #971, opened 2019-11-07
Fix ability to expand nav submenus on large-viewport touch interfaces

### LINKED ISSUE #790: Unable to access a submenu item of the Desktop Horizontal Menu on tablet (>1000px)
On a tablet (1024px for an iPad or 1366px for an iPad Pro for example) when the user clicks on the menu item (or on its icon) which has a submenu, it always loads the parent page. The submenu appears briefly but it's impossible to click on a submenu item.

The submenu works well inside the modal as the user can click on the icon to open the submenu.

westonruter:
In #949 changes were merged to fix #790, the inability to access a submenu item of the Desktop Horizontal Menu on tablet (>1000px). However, I [found](https://github.com/WordPress/twentytwenty/pull/949#issuecomment-551220080) the fix is not working:

image

An alternative approach to adding JS to support touch interfaces is rather to use CSS, specifically media queries for interaction media features ([caniuse](https://caniuse.com/#feat=css-media-interaction), currently 92.57% of browsers globally).

An added benefit of using CSS here is that it will ensure that the nav menus work when JS is turned off in the browser (or it fails to load, or it is an AMP page). Nevertheless, since IE11 doesn't support this media query, perhaps some JS logic should be included to add a `touch-enabled` class name on the `body` and then add CSS rules for them as well (for IE11 users who have JS enabled: <7.43% of global users).

Props to @pierlon who found this solution when devising an AMP-compatible way to make the submenus accessible on large-viewport touch interfaces: https://github.com/ampproject/amp-wp/pull/3696.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1703,6 +1703,18 @@ ul.primary-menu {
 
 }
 
+/* Repeat previous rules for IE11 (when JS enabled for polyfill). */
+body.touch-enabled .primary-menu > li.menu-item-has-children > a {
+	padding-right: 0;
+	margin-right: 2rem;
+}
+
+body.touch-enabled .primary-menu ul li.menu-item-has-children > a {
+	margin-right: 4.5rem;
+	padding-right: 0;
+	width: unset;
+}
+
 /* -------------------------------------------------------------------------- */
 
 /*	5. Menu Modal
--- a/style.css
+++ b/style.css
@@ -1687,6 +1687,22 @@ ul.primary-menu {
 	transform: rotate(180deg);
 }
 
+/* Nav submenu expansion on large-viewport touch screens (e.g. tablets) */
+@media (hover: none) and (pointer: coarse) {
+
+	.primary-menu > li.menu-item-has-children > a {
+		padding-right: 0;
+		margin-right: 2rem;
+	}
+
+	.primary-menu ul li.menu-item-has-children > a {
+		margin-right: 4.5rem;
+		padding-right: 0;
+		width: unset;
+	}
+
+}
+
 /* -------------------------------------------------------------------------- */
 
 /*	5. Menu Modal
--- a/style.css
+++ b/style.css
@@ -1687,8 +1687,12 @@ ul.primary-menu {
 	transform: rotate(180deg);
 }
 
-/* Nav submenu expansion on large-viewport touch screens (e.g. tablets) */
-@media (hover: none) and (pointer: coarse) {
+/*
+ * Enable nav submenu expansion with tapping on arrows on large-viewport
+ * touch interfaces (e.g. tablets or laptops with touch screens).
+ * These rules are supported by all browsers (>IE11) and when JS is disabled.
+ */
+@media (any-pointer: coarse) {
 
 	.primary-menu > li.menu-item-has-children > a {
 		padding-right: 0;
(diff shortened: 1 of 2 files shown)


style.css
```diff
@@ -1687,6 +1687,22 @@ ul.primary-menu {
...
+	}
+
+	.primary-menu ul li.menu-item-has-children > a {
+		margin-right: 4.5rem;
+		padding-right: 0;
+		width: unset;
+	}
+
+}
+
```
westonruter: I am not sure this is the the most logical place for this code.

style.css
```diff
@@ -1687,6 +1687,22 @@ ul.primary-menu {
 	transform: rotate(180deg);
 }
 
+/* Nav submenu expansion on large-viewport touch screens (e.g. tablets) */
+@media (hover: none) and (pointer: coarse) {
```
westonruter: Another option here would be:

```suggestion
@media (any-pointer: coarse) {
```

This should ensure that devices with both touch screens _and_ trackpads (or mice) would still allow tapping to expand submenus. Many laptops today also feature touch screens.

I'm not an expert on these media queries. I relied on https://medium.com/@ferie/detect-a-touch-device-with-only-css-9f8e30fa1134

pattonwebz:
With this change here we rely on CSS and the `interaction media queries` feature but there is no support for that in IE 11 or before. About 1.85% global users are on IE11 according to the [CanIUse stats](https://caniuse.com/#feat=css-media-interaction).

I am interested to hear the though about how best to handle that. I don't think the figure is small enough to ignore it but at the same time I do not want to have to introduce any kind of a JS shim for IE because this PR actually removes some JS code making the menus a little bit easier to work with from a developers perspective.

style.css
westonruter: Done in d5f02f0.

westonruter:
@pattonwebz I've added a polyfill for IE11 in 6f20c02.

pattonwebz:
I see that, I am just trying to boot my laptop into windows so I can give this a live test on IE 11 but otherwise this seems good to me.

After some googling I was not able to find any solution other than a polyfill. It is a shame but that is just what we have to do to keep IE working :man_shrugging:

carolinan:
I am not able to reproduce this JavaScript error.

> In #949 changes were merged to fix #790, the inability to access a submenu item of the Desktop Horizontal Menu on tablet (>1000px). However, I [found](https://github.com/WordPress/twentytwenty/pull/949#issuecomment-551220080) the fix is not working:
> 
> image
> 
> An alternative approach to adding JS to support touch interfaces is rather to use CSS, specifically media queries for interaction media features ([caniuse](https://caniuse.com/#feat=css-media-interaction), currently 92.57% of browsers globally).
> 
> An added benefit of using CSS here is that it will ensure that the nav menus work when JS is turned off in the browser (or it fails to load, or it is an AMP page). Nevertheless, since IE11 doesn't support this media query, perhaps some JS logic should be included to add a `touch-enabled` class name on the `body` and then add CSS rules for them as well (for IE11 users who have JS enabled: <7.43% of global users).
> 
> Props to @pierlon who found this solution when devising an AMP-compatible way to make the submenus accessible on large-viewport touch interfaces: [ampproject/amp-wp#3696](https://github.com/ampproject/amp-wp/pull/3696).

PR #974, opened 2019-11-08
Load the starter content only when within the customizer

### LINKED ISSUE #973: Starter content processed even if not used
I noticed that the way starter content is currently being handled breaks WP-CLI tests when switching Core versions, even though neither the theme nor its starter content would actually be useful to load. See here for some of the breaking tests: https://travis-ci.org/wp-cli/automated-tests/jobs/605262104#L1103-L1109

The main issue is that the entire starter content is loaded, parsed and processed only to add it to the theme support hook. If it will not be used after that, all the entire processing did was just waste resources: not only is memory allocated to construct the array, we also have filesystem checks and translations that will be processed.

As we don't have autoloading in Core yet, I suggest wrapping the starter content in a Closure that only requires the necessary file once the corresponding hook is triggered.

schlessera:
To avoid parsing the starter content on every single page request, we only
add the theme support for the starter content when `is_customizer_preview()` is
true.

Fixes #973

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -119,8 +119,15 @@ function twentytwenty_theme_support() {
 	// Add support for full and wide align images.
 	add_theme_support( 'align-wide' );
 
-	// Adds starter content to highlight the theme on fresh sites.
-	add_theme_support( 'starter-content', twentytwenty_get_starter_content() );
+	/*
+	 * Adds starter content to highlight the theme on fresh sites.
+	 * This is done conditionally to avoid loading the starter content on every
+	 * page load, as it is a one-off operation only needed once in the customizer.
+	 */
+	if ( is_customize_preview() ) {
+		require get_template_directory() . '/inc/starter-content.php';
+		add_theme_support( 'starter-content', twentytwenty_get_starter_content() );
+	}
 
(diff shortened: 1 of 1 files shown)


pattonwebz:
@schlessera Does this also fix all the problems that you were facing in WP-CLI as well caused by this?

schlessera:
@pattonwebz I'm not sure, and it is actually difficult to test if the change is not in a release.

However, I've built a work-around for WP-CLI already to make sure I'm not hit by this when 5.3 is released: https://github.com/wp-cli/wp-cli-tests/pull/72

schlessera:
In your test, can you add a `wp-die()` in the root of the `starter-content` theme and see if it is only hit in the customizer, and does not break outside of it?

pattonwebz:
@schlessera I tested this with xDebug (and again I tested with your wp_die() method) and the file/function is only hit on customizer pages as expected. I haven't found any other admin pages that it fires on but there may be some places or situations in the dashboard that resolve to `true` on `is_customize_preview()` but I have not found any of those in my browsing around just now.

This seems good to me but I will test for another few minutes to make sure I am not missing anything but if I run into no troubles I will merge this in :)

Issue #975, opened 2019-11-09
Input fields have no visible focus in multiple browsers


carolinan:
Users need to be able to see when an input field is selected.
Both to be able to use the fields and to navigate past them, they need to be able to see where they are on the page.

On Windows: 
Below is the comment field when no fields are selected:
(Chrome)

![comment-no-focus](https://user-images.githubusercontent.com/7422055/68528421-baf23c80-02f2-11ea-900e-3758559ffa25.png)

Below is when the comment text area is select, in Chrome:
![comment-selected-chrome](https://user-images.githubusercontent.com/7422055/68528434-f68d0680-02f2-11ea-87a1-c93b49ab34e7.png)

In Firefox, there is no visible outline or similar change:
![comment-selected-firefox](https://user-images.githubusercontent.com/7422055/68528472-68655000-02f3-11ea-9591-bb08384c86e1.png)

In Edge, there is no visible outline or similar change:
![comment-selected-edge](https://user-images.githubusercontent.com/7422055/68528523-1113af80-02f4-11ea-9d32-da0a821eaf49.png)

In IE 11, there is no visible outline or similar change:
![comment-selected-ie11](https://user-images.githubusercontent.com/7422055/68528556-ae6ee380-02f4-11ea-8b4c-4faefb134370.png)

andersnoren:
Fixed by #976.

PR #976, opened 2019-11-09
Update outline for input fields

carolinan:
Adds a very basic 1px dotted outline for input fields when they are focused.

This outline will be inconsistent for Chrome users. Ideally the outline should be consistent across the entire theme, not only for input fields, and independent on browsers.

Quick fix for https://github.com/WordPress/twentytwenty/issues/975

### DIFF
--- a/style-rtl.css
+++ b/style-rtl.css
@@ -826,6 +826,26 @@ input[type="submit"]:hover {
 	text-decoration: underline;
 }
 
+input[type="text"]:focus,
+input[type="email"]:focus,
+input[type="url"]:focus,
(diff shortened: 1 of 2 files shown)


carolinan:
Firefox before:
The comment form text area has no outline, border, shadow or similar decoration change when focused

And after:
The comment form text area has a one pixel, dark, dotted outline when focused

carolinan:
With border color change:

The updated version shows the border color with the current text color when focused

And outline on search form:
The search modal input field has a one pixel dotted border when focused

PR #981, opened 2019-11-09
doctype uppercase to lowercase

bappi:
There's no reason for the uppercase doctype. It was discussed and committed to _s repo also

### DIFF
--- a/header.php
+++ b/header.php
@@ -9,7 +9,7 @@
  * @since 1.0.0
  */
 
-?><!DOCTYPE html>
+?><!doctype html>
 
 <html class="no-js" <?php language_attributes(); ?>>
 


pattonwebz:
Hey @bappi thanks for opening this PR. I looked into the discussion on _s repo you mentioned and followed up on a few connected issues/merges elsewhere and still I am not sure if there are many compelling reasons to change this. 

It is case-insensitive at specification level and the primary reasons I found for swapping to lowercase are about possible compression improvements - theoretically the lowercase characters are more common throughout the majority of documents and thus there is potential for fractions of bytes to be saved.

I don't see that there's a real benefit to swapping this or any major downside to keeping uppercase. Are there other things I am missing about the swap?

bappi:
There aren't any major things about it, the thing is it's becoming more common to use lowercase doctype in all other places.  

So we should consider this to keep the flow.

pattonwebz:
Understood, thanks for the explanation. I was worried I had missed something when I looked it up.

Looking at some other themes I see that it is becoming more common to user lowercase. I do not have a preference either way. This could be merged if nobody objects.

CC: @ianbelanger79 @andersnoren @carolinan for any farther discussions if needed.

nielslange:
Apparently, `DOCTYPE` was case-sensitive up to XHTML 1.1, and is case-insensitive since HTML5. That said, I created a small list to highlight which site is following which convention:

<table>
<tr>
<td>https://apple.com/</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://cnn.com</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://developer.mozilla.org/</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://github.com/</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://hackerone.com/</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://medium.com/</td>
<td><code>&lt;!doctype html></code></td>
</tr>
<tr>
<td>https://microsoft.com/</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://techcrunch.com/</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://wordpress.com</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
<tr>
<td>https://wordpress.org/</td>
<td><code>&lt;!DOCTYPE html></code></td>
</tr>
</table>

Based on these stats and due to legacy reasons, personally I would stick to <code>&lt;!DOCTYPE html></code>.

ianbelanger79:
I agree with @nielslange, we should leave it as-is for now. Closing this PR as `wontfix`

PR #955, opened 2019-11-06
Take 2 - Update #828 to address the starter_content vanishing on mobile

garrett-eclipse:
Update #828 to address the starter_content vanishing on mobile with the default menu setup. This will remove the expanded desktop menu and place all items in the primary menu.

This replaced #876 

Slack thread from WP 5.3 RC4 - https://wordpress.slack.com/archives/C02RQBWTW/p1572998276403100

PR #987, opened 2019-11-10
Garrett eclipse starter content mobile menu

pattonwebz:
This pulls in PR #955 which fixes a problem with some content not being reachable due to menu configurations. It also replicates the same menu entries in the expanded location to demonstrate it's abilities.


garrett-eclipse:
Thanks @pattonwebz much appreciated

Issue #991, opened 2019-11-10
NEW ISSUES SHOULD BE CREATED ON TRAC


pattonwebz:
From this point on any new issues should be opened on the core trac instead of at GitHub to save them needing to be transitioned over there manually at a later point.

Please prefix tickets with `Twenty Twenty:` so that they are easy to find over there.

https://core.trac.wordpress.org

ottok:
Why do you link to trac main page here? Please link to the list of twentytwenty issues directly. Thanks!

The commit https://github.com/WordPress/twentytwenty/commit/1025b5284f2927ed0b92f964b30243eab4d3b5e8#diff-3d9639f99b2cd6bda90b892bea090bef broke this plugin for PHP composer users, as there is a dotfile now shipped with the plugin and composer does not handle well those cases. This is a similar bug as when akismet started shipping a .htaccess file in May 2017.

```
# composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Package operations: 4 installs, 0 updates, 0 removals
  - Installing johnpbloch/wordpress-core (5.3.0): Loading from cache
  - Installing wpackagist-plugin/auto-post-thumbnail (3.6.0): Downloading (100%)         
  - Installing wpackagist-plugin/google-analytics-dashboard-for-wp (5.3.9): Downloading (100%)         
  - Installing wpackagist-theme/twentynineteen (1.4): Downloading (100%)         
Generating autoload files
> WordPress\Installer::symlinkWPContent
Script WordPress\Installer::symlinkWPContent handling the post-install-cmd event terminated with an exception

                                                                                               
  [ErrorException]                                                                             
  rmdir(/data/wordpress/htdocs/wordpress/wp-content/themes/twentytwenty): Directory not empty                                                                                            
```

pattonwebz:
Hey @ottok, we can certainly update the issue template so it links directly to the tickets, do you think also linking directly to the new ticket page would be good too?

As for composer problems with dotfiles in the directory - that is unfortunate and I will look into that and see what can be done, thank you for reporting.

ottok:
No, just link to the issue overview. You want users to skim existing issues before going to file new ones.

For the composer issue I wrote a local fix to our copy of the general WordPress/Install.php used in in many gitified WordPress templates: https://github.com/Seravo/wordpress/commit/a42412cdb69074ab3b8645df6150adfae4cdd0c0

PR #993, opened 2019-11-11
Update issue templates

nielslange:
@pattonwebz: I created this issue template, in case someone missed your issue #991. What happens is, that if one creates a new issue, that person will see your text from #991.

### DIFF
--- a/.github/ISSUE_TEMPLATE/create-new-issues-on-trac.md
+++ b/.github/ISSUE_TEMPLATE/create-new-issues-on-trac.md
@@ -1,10 +1,5 @@
 ---
 name: Create new issues on Trac
-about: Create new issues on Trac
-title: ''
-labels: ''
-assignees: ''
-
 ---
 
 # NEW ISSUES SHOULD BE CREATED ON TRAC
--- a/.github/ISSUE_TEMPLATE/create-new-issues-on-trac.md
+++ b/.github/ISSUE_TEMPLATE/create-new-issues-on-trac.md
@@ -1,0 +1,18 @@
+---
+name: Create new issues on Trac
+about: Create new issues on Trac
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+# NEW ISSUES SHOULD BE CREATED ON TRAC
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/create-new-issues-on-trac.md
```diff
@@ -0,0 +1,18 @@
...
+labels: ''
+assignees: ''
+
+---
+
+# NEW ISSUES SHOULD BE CREATED ON TRAC
+
+From this point on any new issues should be opened on the core trac instead of at GitHub to save them needing to be transitioned over there manually at a later point.
+
+Please prefix tickets with Twenty Twenty: so that they are easy to find over there.
```
Aurorum: Probably better to make this a bit more clear like this:

```suggestion
Please prefix tickets with `Twenty Twenty:` so that they are easy to find over there.
```

.github/ISSUE_TEMPLATE/create-new-issues-on-trac.md
```diff
@@ -0,0 +1,18 @@
+---
+name: Create new issues on Trac
+about: Create new issues on Trac
```
Aurorum: This appearing twice does seem a bit bizarre, but I guess it doesn't really matter.

Screenshot_20191111-083835

nielslange:
> Would it be better/possible to just close issues instead?

That's also an option, but this option might backfire, if potential contributors don't know why their issues got closed. This might send out wrong signals.

.github/ISSUE_TEMPLATE/create-new-issues-on-trac.md
nielslange: Fixed!

pattonwebz:
This looks like a good idea to me, much cleaner option :)

nielslange:
> This looks like a good idea to me, much cleaner option :)

Of course, once WordPress 5.3 got released, we can then achieve this repo, similar to https://github.com/WordPress/twentynineteen, so that no more issues can be created in the first place. :)

Issue #964, opened 2019-11-06
Selecting a Background Color preset for the Group block should automatically change the text color


yannickiki:
Some blocks don't have text color settings (Quote, List, Table...). If the user adds one of these blocks inside a group and selects the Primary, Accent or Secondary Color as the Background Color, the text is unreadable (and not AA compliant). It also doesn't work with custom colors.

Like Twenty Nineteen, Twenty Twenty should at least define the text color in accordance with the background color presets.

Group without Background Color | Group with Primary Background Color
-- | --
![no-background](https://user-images.githubusercontent.com/54663751/68327150-17741280-00cd-11ea-8df2-0f5b2c750acc.png) | ![background](https://user-images.githubusercontent.com/54663751/68327176-21961100-00cd-11ea-9a7d-a6f21d9c1fa8.png)

pattonwebz:
Hey @yannickiki I merged some code which I think resolves this as best as we can right now. If other issues are noted could we report them via trac so we can work on them after release. Thank you!

PR #995, opened 2019-11-11
Add a11y friendly default text-colors when selecting a background-color.

allancole:
**Before:**

**After:**

#### Related Tickets 

Fixes: #964

### DIFF
--- a/style-rtl.css
+++ b/style-rtl.css
@@ -2372,6 +2372,10 @@ h2.entry-title {
 
 .post-meta-wrapper {
 	margin-top: 2rem;
+	margin-left: auto;
+	margin-right: auto;
+	max-width: 58rem;
+	width: calc(100% - 4rem);
 }
 
 .post-meta {
@@ -2532,6 +2536,10 @@ h2.entry-title {
 
 .author-bio {
 	margin-top: 4rem;
+	margin-left: auto;
+	margin-right: auto;
+	max-width: 58rem;
+	width: calc(100% - 4rem);
 }
 
 .hide-avatars .author-bio {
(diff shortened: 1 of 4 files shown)


joyously:
Don't the color settings change in the palette when you change the colors in the Customizer?
So setting the text color is only good for the defaults?
Wouldn't it be better to make the text color work regardless of background color?

allancole:
Group Blocks and any block that uses any of the `.has-background-xxxxxxx-color` utility classes are also fixed with this PR. Here’s a gif of it working with this fix:

mapk:
I just tested this in FF 70.0.1 and am seeing the a11y warning still. Is this right?

contrast 2019-11-11 14_32_50

mapk:
So double checking the contrast on these color combinations:

- RED (#cd2653) & WHITE (#FFF) – 5.25 contrast ratio
- BLACK (#000) & TAN (#f5efe0) – 18.30 contrast ratio
- DARK GRAY (#6d6d6d) & WHITE (#FFF) – 5.17 contrast ratio

Looks like something may be off with the contrast warning in Gutenberg?

joyously:
I tried it with changed colors.
Peek 2019-11-11 16-48

style-rtl.css
```diff
@@ -2664,6 +2666,7 @@ h2.entry-title {
 
 :root .has-secondary-background-color {
 	background-color: #6d6d6d;
+	color: #FFF;
```
audrasjb: Coding standards: should be `#fff` for consistency.

audrasjb:
Hi,
For what it worth, I tested the PR on my side, and it looks good to me.
Changing background color is automatically changing the font color when needed.

audrasjb:
5e7ff2db531ab3c9150268adb25fc914

Issue #773, opened 2019-10-05
RTL: text aligned blocks are reversed, some aligned blocks reversed


joyously:
The block styles loaded on the front end contain the correct rules for `.has-text-align-left` and `.has-text-align-right`, but the theme styles override with the wrong direction. These should just be removed.

The core blocks aligned left, right, center in LTR:
![left-right-center-LTR](https://user-images.githubusercontent.com/7957680/66261714-141cfc80-e798-11e9-86bc-7a819e933e20.jpg)

Here is the same post in RTL:
![left-right-center-RTL](https://user-images.githubusercontent.com/7957680/66261717-18e1b080-e798-11e9-9d36-61415d5545a4.jpg)

The problem with the Media/Text blocks is due to CSS in the block styles. I opened an issue for it on [GB repo](https://github.com/WordPress/gutenberg/issues/17787). Here is how it looks on RTL with one CSS rule disabled:
![media-text-with-smaller-media-RTL](https://user-images.githubusercontent.com/7957680/66261743-7bd34780-e798-11e9-9dfb-ee9bac0e834d.jpg)

-----------------------------
Here is the test content:
```
<!-- wp:heading {"align":"left","level":4} -->
<h4 class="has-text-align-left">h4 left</h4>
<!-- /wp:heading -->

<!-- wp:heading {"align":"right","level":4} -->
<h4 class="has-text-align-right">h4 right</h4>
<!-- /wp:heading -->

<!-- wp:heading {"align":"center","level":4} -->
<h4 class="has-text-align-center">h4 center</h4>
<!-- /wp:heading -->

<!-- wp:paragraph {"align":"left"} -->
<p class="has-text-align-left">Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Donec mollis. Quisque convallis libero in sapien pharetra tincidunt. Aliquam elit ante, malesuada id, tempor eu, gravida id, odio. </p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"align":"right"} -->
<p class="has-text-align-right">Maecenas suscipit, risus et eleifend imperdiet, nisi orci ullamcorper massa, et adipiscing orci velit quis magna. Praesent sit amet ligula id orci venenatis auctor. Phasellus porttitor, metus non tincidunt dapibus, orci pede pretium neque, sit amet adipiscing ipsum lectus et libero. Aenean bibendum.</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"align":"center"} -->
[...]

PR #998, opened 2019-11-12
Excluding pullquote block from width resize rules

allancole:
Fixes the issue mentioned here after merging the new responsive width logic: 
https://github.com/WordPress/twentytwenty/pull/701#issuecomment-552655172

After fix:


joyously:
This still does not fix the issues mentioned in #773

PR #1001, opened 2019-11-12
Remove .section-inner.thin from the cover template

andersnoren:
Removes the `.section-inner.thin` classes from the `.post-inner` element on the cover template, matching the structure used on `content.php` and fixing the alignment issue seen on http://2020.wordpress.net/about/.


carolinan:
LGTM

PR #1005, opened 2019-11-12
Remove the 0 left-margin on some recent post blocks, uses auto instead

pattonwebz:
fixes item from doc: _The latest posts widget block isn’t aligned correctly_


jasmussen:
Tests well for me, ship it:

latestposts

I can't approve here, but rest assured I'm here to help and I did in fact test :)

